PR #20, opened 2013-05-06
Support for CDI environments that do not support injection into servlet context listeners

Blimster:
I want to use Vaadin with CDI in a servlet container like Tomcat 7.

To make CDI available in Tomcat, I use the weld-servlet, which not supports injection into listeners. Unfortunately, the vaadin-cdi-integration is using a context listener to deploy the vaadin servlet.

In my commit, the ContextDeployer is only executed if the injection of the BeanManager was successful. In the case you have to specify the Vaadin servlet on the CDIUIProvider in the web.xml.

In the CDIUIProvider, if the BeanManager is not found by the name "java:comp/BeanManager" then it also tries to lookup with the name "java:comp/env/BeanManager" (Tomcat provides resource under "java:comp/env/". You must also specify the BeanManager as a resource in the context.xml and in the web.xml as described at http://docs.jboss.org/weld/reference/1.0.0/en-US/html/environments.html#d0e4998.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/20)

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIUIProvider.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIUIProvider.java
@@ -196,12 +196,23 @@ public class CDIUIProvider extends DefaultUIProvider implements Serializable {
         if (beanManager == null) {
             // as the CDIUIProvider is not injected, need to use JNDI lookup
             try {
-                InitialContext initialContext = new InitialContext();
+            	InitialContext initialContext = new InitialContext();
                 beanManager = (BeanManager) initialContext
                         .lookup("java:comp/BeanManager");
             } catch (NamingException e) {
-                getLogger().severe("Could not get BeanManager through JNDI");
-                beanManager = null;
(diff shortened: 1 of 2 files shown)


vaadin-cdi/src/main/java/com/vaadin/cdi/CDIUIProvider.java
```diff
@@ -196,12 +196,23 @@ private BeanManager getBeanManager() {
...
+
+                try
+                {
+	            	InitialContext initialContext = new InitialContext();
+	                beanManager = (BeanManager) initialContext
+	                        .lookup("java:comp/env/BeanManager");
+                }
+                catch(NamingException e2)
+                {
+                    getLogger().info("Could not get BeanManager through JNDI");
```
doggy-dev: I think you should keep severe level here

Blimster: Yes, you are right.

But after over 3 year, is this pull request still relevant? I didn't noticed any interest for this pull request until your comment.

hesara:
Relevant fix already in master (using CDIUtil.lookupBeanManager()).

PR #180, opened 2016-10-25
Make the CDI plugin more customizable

nahkh:
You can now override the default convention behaviour by activating an alternative implementation for the Conventions interface.
Made CDIViewProvider#getViewBean and UIViewProvider#getUIBeanWithMapping protected to allow overriding.
Refactored the CDIUIProvider and CDIViewProvider such that the plugin's default annotations are only accessed through a class implementing the Conventions interface. This way client applications can provide their own specialised annotations and/or mapping conventions simply by implementing Conventions as an alternative CDI bean and activating it within their application.

---

This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/180)


pleku:
Hello. Sorry for no progress on this; but we've finally considered the changes and determined that due to _no demand_ for these changes, we're not going to include the changes for now.

If there is something in this PR that you'd like to get into the CDI add-on, please open an enhancement issue and propose the solution(s) there before making a changeset of this magnitude. Thanks

PR #183, opened 2016-11-20
some test related fixes, and managed arquillian test profiles for glassfish, and wildfly

kumm:
changes:
- Multiple tests tried to keep synchronized with client side using Thread.sleep. I've introduced a simple way to wait for vaadin client side using testbench javascript hook. Just to be clear: no testbench dependencies introduced. The hook exists in vaadin client.
-  Converted all tests to run as client. There were many static member accesses to the deployed classes ( mostly counters ) by the test. The main reason is embedded wildfly don't work with vaadin.  For some reason wildfly arquillian use a servlet even in embedded mode. Vaadin cdi deploys the servlet at /* knocking out the wildfly arquillian servlet. Maybe there is a solution, but never mind. Manged tests are always better than embedded. 
- Added glassfish, and wildfly managed arquillian test profiles. Some minor arquillian, and drone version upgrades also happened. Didn't touch tomee profile, it's still embedded.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/183)

### DIFF
--- a/vaadin-cdi/src/test/java/com/vaadin/cdi/MultipleAccessIsolationTest.java
+++ b/vaadin-cdi/src/test/java/com/vaadin/cdi/MultipleAccessIsolationTest.java
@@ -26,7 +26,7 @@ public class MultipleAccessIsolationTest extends
         resetCounts();
     }
 
-    @Deployment(name = "concurrentAccess")
+    @Deployment(name = "concurrentAccess", testable = false)
     public static WebArchive alternativeAndActiveWithSamePath() {
         return ArchiveProvider.createWebArchive("concurrentAccess",
                 ConcurrentUI.class);
--- a/vaadin-cdi/src/test/java/com/vaadin/cdi/MultipleAccessIsolationTest.java
+++ b/vaadin-cdi/src/test/java/com/vaadin/cdi/MultipleAccessIsolationTest.java
@@ -41,17 +41,19 @@ public class MultipleAccessIsolationTest extends
         assertThat(getCount(ConcurrentUI.CONSTRUCT_COUNT), is(1));
         firstWindow.findElement(By.id(ConcurrentUI.OPEN_WINDOW)).click();
         waitForNumberOfWindowsToEqual(2);
-        Thread.sleep(500);
         List<String> handles = new ArrayList<String>(
                 firstWindow.getWindowHandles());
+        // wait for second window fully loads
+        firstWindow.switchTo().window(handles.get(1));
+        (new WebDriverWait(firstWindow, 15)).until(ExpectedConditions
+                .presenceOfElementLocated(LABEL));
+
         firstWindow.switchTo().window(handles.get(0));
         firstWindow.findElement(By.id(ConcurrentUI.COUNTER_BUTTON)).click();
         firstWindow.findElement(By.id(ConcurrentUI.COUNTER_BUTTON)).click();
         firstWindow.findElement(By.id(ConcurrentUI.COUNTER_BUTTON)).click();
         waitForValue(By.id(ConcurrentUI.COUNTER_LABEL), 3);
         firstWindow.switchTo().window(handles.get(1));
-        (new WebDriverWait(firstWindow, 15)).until(ExpectedConditions
-                .presenceOfElementLocated(LABEL));
         // the second window was loaded before the clicks on the first one so
         // the counter is always 0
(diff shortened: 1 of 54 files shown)


vaadin-cdi/src/test/java/com/vaadin/cdi/uis/RootUI.java
```diff
@@ -16,28 +16,27 @@
...
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+
 /**
  */
 @CDIUI("")
 public class RootUI extends UI {
-    private final static AtomicInteger COUNTER = new AtomicInteger(0);
-    private int clickCount;
+    public static final String CONSTRUCT_KEY = "RootUIConstruct";
```
hesara: Is there a reason this key is named differently from the rest?

vaadin-cdi/src/test/java/com/vaadin/cdi/internal/MyBean.java
```diff
@@ -10,7 +9,6 @@
     private final int id = counter++;
 
     public MyBean() {
-        System.out.println(UI.getCurrent());
```
hesara: It would be good to understand why this causes an NPE, but that doesn't block merging these changes.

vaadin-cdi/src/test/java/com/vaadin/cdi/MultipleAccessIsolationTest.java
```diff
@@ -41,17 +41,19 @@ public void testConcurrentAccess() throws IOException,
...
         waitForNumberOfWindowsToEqual(2);
-        Thread.sleep(500);
         List<String> handles = new ArrayList<String>(
                 firstWindow.getWindowHandles());
+        // wait for second window fully loads
+        firstWindow.switchTo().window(handles.get(1));
+        (new WebDriverWait(firstWindow, 15)).until(ExpectedConditions
+                .presenceOfElementLocated(LABEL));
+
         firstWindow.switchTo().window(handles.get(0));
```
hesara: (No wait needed after any of the other window.switchTo() calls? Is everything already loaded in them? I assume so, haven't dug deeper.)

hesara:
See comments from GH review tool. Also, the change uses * imports which we have avoided, but that won't prevent merging these changes - we can also reorganize imports for the whole project sometime in the future.

---

Reviewed 53 of 53 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/183#-:-KaBVqyes4KQHgGb2Fd_:blpsl1e)*

vaadin-cdi/src/test/java/com/vaadin/cdi/uis/RootUI.java
kumm: I've missed to rename it when changed everywhere to '_COUNT'. fixing.

vaadin-cdi/src/test/java/com/vaadin/cdi/internal/MyBean.java
kumm: I've found this: https://java.net/jira/browse/GLASSFISH-21242.
According to this [changelog](https://payara.gitbooks.io/payara-server/content/v/162/release-notes/release-notes-144.html) payara fixed it. I can switch to payara if you think it worth.

vaadin-cdi/src/test/java/com/vaadin/cdi/MultipleAccessIsolationTest.java
kumm: I see that all switchTo switches to a window already loaded.
line 51: window 0 loaded because openWindow on line 4 waits to label. Click on line 42 does not trigger any server side event listener.
line 56: window 1 loaded because of wait on line 48-49

kumm:
rebased onto master, and removed reindeer theme from a test UI

hesara:
:lgtm:

---

Reviewed 2 of 2 files at r2, 3 of 3 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/183#-:-KbPMHUeZBc9v98OliwS:bnfp4nl)*

hesara:
Sorry for the delay with this - I was out of the office for more than a week.
Tested quickly with all three profiles (tomee, wildfly, glassfish).

PR #184, opened 2016-11-27
UIBean factor out, more fixes for contextual instance destroy

kumm:
This pull request is based on #183 . Starting at commit b2291b3.
Contents:
- a test, and fix for #176 ( first two commit is rebased from pull request #178 )
- a refactor to drop UIBean concept. I found this concept makes hard to properly destroy contextual instances.
- tests, and fixes to more issues related to destroying contextual instances. #169 #156 #158 .

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/184)

Issue #176, opened 2016-10-04
UIScopedContext leaks closed UIs


kumm:
All closed UI is referenced from UIScopedContext storageMap until http session invalidation.

Simple to reproduce, just build a helllo-world vaadin app with one CDIUI.
Run it, and refresh the page in the browser. Every time you refresh, an UI with closing=true leaked from UIScopedContext. - I know i can use PreserveOnRefresh.
But same result when opening the UI in a new tab, closing it, and waiting 3 heartbeatInterval.

I think UIScopedContext should free the UI after it's closed.

Here is a screenshot from jvisualvm. [![vaadincdijvm1.png](https://s11.postimg.org/luw1u4yfn/vaadincdijvm1.png)](https://postimg.org/image/gw8jflumn/)

Tested with CDi addon 1.0.3, and vaadin 7.7.2.

kumm:
After some investigation in the source code i can explain the intended behavior.
Closed UIs of a vaadinsession should destroy, and free form UIContext.storageMap on the first viewChange event after a five second timeout.
But actually it does not work.

m--stein:
Same problem here.
Any chance that the fix from kumm will be integrated soon?

doggy-dev:
This seems related to #169

kumm:
@doggy-dev : There are multiple problems in this add-on with destroying contextual instances. #169 is about viewcontext, so not related. This issue is about leaking UIs. 
The pull request #178 for this issue fixes destroying of UIs, not Views ( not even UIScoped beans ).
But we can say issue #155 is closely related, and fixed by the pull request.

zeitiger:
I have a similar problem with `com.vaadin.cdi.internal.AbstractVaadinContext#storageMap` because in application `storageMap.remove(...)` is not called. 
The pull request #184 doesn't work for me :-(

kumm:
@zeitiger can you share the details?
How do you test it exactly? (cleanup after 5 seconds on the next view navigation should work.)
Can you share a simple project (or integration test) to reproduce the issue?

zeitiger:
Currently I still make my research and try to understand what exactly happing. The 5 seconds clean up doesn't work for me. New session could be created within my application (incognito browser or logout), but the session will never clean up. It is quite hard to track down why something doesn't happen.

zeitiger:
[...]

PR #193, opened 2017-01-29
fix for closed UI cleanup

kumm:
It's a fix for #176.
Part 1 of #184

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/193)

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/AbstractVaadinContext.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/AbstractVaadinContext.java
@@ -26,7 +26,6 @@ import java.util.logging.Logger;
 
 import javax.enterprise.context.spi.Contextual;
 import javax.enterprise.context.spi.CreationalContext;
-import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.BeanManager;
 
 import org.apache.deltaspike.core.util.context.AbstractContext;
@@ -34,7 +33,6 @@ import org.apache.deltaspike.core.util.context.ContextualStorage;
 
 import com.vaadin.cdi.internal.AbstractVaadinContext.SessionData.UIData;
 import com.vaadin.server.VaadinSession;
-import com.vaadin.ui.UI;
 
 /**
  * UIScopedContext is the context for @UIScoped beans.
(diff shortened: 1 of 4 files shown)


hesara:
The change itself looks ok, but some comments about tests where you could perhaps comment on the way things are done, and this could be picked/rebased on top of master as the ShiroTest change was already merged there.

---

Reviewed 5 of 7 files at r1.
Review status: 5 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/UIDestroyTest.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/cdi/193#-KbiG5I0_XiqgsxcnTv5:-KbiG5I0_XiqgsxcnTv6:blue0wr) ([raw file](https://github.com/vaadin/cdi/blob/b1b322eb8fd88bffacd79f6257f5d81735d4396c/vaadin-cdi/src/test/java/com/vaadin/cdi/UIDestroyTest.java#L41)):*
> ```Java
>         assertDestroyCount(0);
> 
>         Thread.sleep(5000); //AbstractVaadinContext.CLEANUP_DELAY
> ```

if this is the exact delay, this test might be somewhat fragile (not sure if an issue)

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/uis/DestroyUI.java, line 71 at r1](https://reviewable.io:443/reviews/vaadin/cdi/193#-KbiGKb3BKPVzuVBM5u7:-KbiGKb3BKPVzuVBM5u8:b-1o9u7v) ([raw file](https://github.com/vaadin/cdi/blob/b1b322eb8fd88bffacd79f6257f5d81735d4396c/vaadin-cdi/src/test/java/com/vaadin/cdi/uis/DestroyUI.java#L71)):*
> ```Java
>             @Override
>             public void buttonClick(Button.ClickEvent event) {
>                 final Navigator navigator = new Navigator(DestroyUI.this, new ViewDisplay() {
> ```

unusual pattern for using the navigator (usually set up for the UI once at initialization), so not quite sure how well this test corresponds to a typical use case - although this might correspond more or less to a "set up navigator at login" case

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/193#-:-KbiHFMzxPuVKfkyMTeM:b-kc1e79)*

kumm:
> if this is the exact delay, this test might be somewhat fragile (not sure if an issue)
Theoretically you are right. If opening the new UI, and processing the navigation click request executes under less than a millisecond, than it should fail. Who knows, cpus are getting faster and faster :)
Going to increment it.
What do you think increasing the visibilitiy of AbstractVaadinContext.CLEANUP_DELAY for test purpose would be better then duplicating the value?

kumm:
> unusual pattern for using the navigator (usually set up for the UI once at initialization), so not quite sure how well this test corresponds to a typical use case - although this might correspond more or less to a "set up navigator at login" case

Indeed it's a bit unusual, but i have a reason.
Creating a navigator in init causes a navigation in com.vaadin.ui.UI#doInit. (Kickstarting the navigation)
I want to avoid any other navigation to be sure my navigation causes the cleanup.
Additionally i would have to provide a root view, or an error view to avoid the IllegalStateException would be caused by this kickstart navigation.

By the way this pattern can be found in other test UIs already (NoViewProviderNavigationUI, InstrumentedUI)

kumm:
rebased onto master

hesara:
:lgtm: - I suppose we can merge this as is

---

Reviewed 2 of 7 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/UIDestroyTest.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/cdi/193#-KbiG5I0_XiqgsxcnTv5:-KboKsHZqW1Ir-s3RndO:bgji2ij) ([raw file](https://github.com/vaadin/cdi/blob/b1b322eb8fd88bffacd79f6257f5d81735d4396c/vaadin-cdi/src/test/java/com/vaadin/cdi/UIDestroyTest.java#L41)):*

if this is the exact delay, this test might be somewhat fragile (not sure if an issue)
</blockquote></details>

probably not too likely to fail the test; the value could also be made public but ok as-is

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/193#-:-KboLEeUn5fMzcUPxCcQ:b-6zh3wa)*

PR #194, opened 2017-01-31
refactor to drop UIBean concept, and fix bean destroy

### LINKED ISSUE #156: @PreDestroy not invoked  on view scoped beans when changing view
**Originally by _jaroslaw_przygodzki_**

---

When  toggling between views preDestroy method is never called on view scoped view.

Affected Version: 1.0.2
Reproducible: Always
Test case: see pull request on GitHub

---

 Imported from https://dev.vaadin.com/ issue #16699

kumm:
After drop of UIBeans it is possible to call deltaspike AbstractContext.destroyAllActive.
Should fix #156, #169

Part 2 of  #184

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/194)


hesara:
Partial review with comments about some potential concerns, not sure if real issues. I haven't been looking at how this works in a long time, so probably still need to take another look at some of the changes.

---

Reviewed 14 of 16 files at r1.
Review status: 14 of 16 files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/AbstractVaadinContext.java, line 50 at r1](https://reviewable.io:443/reviews/vaadin/cdi/194#-KbuKw1L_KAP-oZD_Jdf:-KbuKw1L_KAP-oZD_Jdg:be91qmu) ([raw file](https://github.com/vaadin/cdi/blob/b3e0da07d9e861b0c21e4c3786ba8820de6e65a9/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/AbstractVaadinContext.java#L50)):*
> ```Java
>     private Map<Long, SessionData> storageMap = new ConcurrentHashMap<Long, SessionData>();
> 
>     public static class StorageKey {
> ```

Should this be serializable for clustering etc.? Or is it not necessary due to life cycle or are there other, more fundamental issues preventing serialization anyway?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/UIScopedContext.java, line 51 at r1](https://reviewable.io:443/reviews/vaadin/cdi/194#-KbuObpdG2SlrFMi9yJk:-KbuObpdG2SlrFMi9yJl:bb64v0e) ([raw file](https://github.com/vaadin/cdi/blob/b3e0da07d9e861b0c21e4c3786ba8820de6e65a9/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/UIScopedContext.java#L51)):*
> ```Java
>             return CurrentInstance.get(StorageKey.class);
>         } else {
>             return new StorageKey(UI.getCurrent().getUIId());
> ```

Probably not an issue, but writing down a potential concern based on vague past recollections: with some CDI implementations if the UI itself is a normal scoped bean and the implementation uses a proxy for it, could it be that UI.getCurrent() methods don't work at this point as its scope hasn't been resolved yet. Might well not be an issue, or only has been an issue with some older versions of some CDI implementations.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinContextualStorage.java, line 35 at r1](https://reviewable.io:443/reviews/vaadin/cdi/194#-KbuNQYCFzz6kgFVcC9F:-KbuNQYDmFJA1Bd54x3H:bm1qk9u) ([raw file](https://github.com/vaadin/cdi/blob/b3e0da07d9e861b0c21e4c3786ba8820de6e65a9/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinContextualStorage.java#L35)):*
> ```Java
>     @Override
>     public Contextual<?> getBean(Object beanKey) {
>         if (beanKey instanceof String) {
> ```

Strings are always keys here and keys are always strings? Or could there be some case where this ends up in the wrong branch?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/194#-:-KbuPk_TY-DNP1Lus9Ve:bqwji4v)*

kumm:
> Should this be serializable for clustering etc.? Or is it not necessary due to life cycle or are there other, more fundamental issues preventing serialization anyway?

I have some experimental work to support clustering (via session replication). I can say there are fundamental things need to be changed.

> Probably not an issue, but writing down a potential concern based on vague past recollections: with some CDI implementations if the UI itself is a normal scoped bean and the implementation uses a proxy for it, could it be that UI.getCurrent() methods don't work at this point as its scope hasn't been resolved yet. Might well not be an issue, or only has been an issue with some older versions of some CDI implementations.

You are right. I've just tested with a normal scoped UI, and got an exception. UI.setCurrent called too late on UI creation, and the proxy fails to resolve the scope when setSession called on the ui (proxy) instance. I will try to fix it.

But to be a bit destructive:  support for normal scoped UIs is in a non defined state. There are no tests(will create one for the fix), generates warning on deploy, violates core framework asserts, and even javadoc of NormalUIScoped states you cannot use it on Vaadin Components. Furthermore it complicates UI cleanup with a global queue. I don't think anybody was brave enough to put such an UI in production. Can't we let it go some day?

> Strings are always keys here and keys are always strings? Or could there be some case where this ends up in the wrong branch?

As i understand keys are come from getBeanKey. It gives the bean itself ( a bean can't be a String ) or the passivation id (it is always a String). So String means the bean was passivationCapable.
getBean called from deltaspike AbstractContext.destroyAllActive with the entry key. The entry created with key from getBeanKey in method ContextualStorage.createContextualInstance.

kumm:
After some investigation i see, i have totally broke handling of normal scoped UIs. 
I think, there is no simple solution. 
The fix would be to reintroduce the uibean concept (or something similar), and start again with an other approach to properly destroy all beans.

kumm:
Thanks for pointing out the problem, please ignore this PR.  At least now i know why uibean needed. I have found a way to keep it and will send as an other PR.

PR #203, opened 2017-04-13
Add wildcard mapping support for UIs

### LINKED ISSUE #202: Wildcard @CDIUI mapping
I need my UI to handle requests from all URLs, i.e.

- `/`
- `/foo`
- `/foo/bar`

which works with `@WebServlet(urlPatterns = "/*")` but not with `@CDIUI("/*")`, nor `@CDIUI("*")`.

How can I achieve this with `@CDIUI`?

P.S. I'm using the new FW 8 History API to push/pop states (instead of using URI fragments), so I really need this to make it work with CDI 🎉

heruan:
This add support for `@CDIUI("*")` and `@CDIUI("foo*")` to handle multiple paths with the same `UI`, e.g.

```java
@CDIUI("*")
public class MyUI extends UI {
  // ...
}
```
will serve `/`, `/foo`, `/foo/bar`.

This is needed when using the new FW 8 History API to push/pop states to the URI.

Closes #202.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/203)

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIUIProvider.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIUIProvider.java
@@ -166,7 +166,9 @@ public class CDIUIProvider extends DefaultUIProvider implements Serializable {
                 if (beanClass.isAnnotationPresent(CDIUI.class)) {
                     String computedMapping = Conventions
                             .deriveMappingForUI(beanClass);
-                    if (mapping.equals(computedMapping)) {
+                    int wildcardIndex = computedMapping.lastIndexOf("*");
+                    if (mapping.equals(computedMapping) || (wildcardIndex >= 0
+                            && mapping.startsWith(computedMapping.substring(0, wildcardIndex)))) {
                         return bean;
                     }
                 }
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/AnnotationUtil.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/AnnotationUtil.java
@@ -61,7 +61,7 @@ public class AnnotationUtil {
             }
 
             String path = Conventions.deriveMappingForUI(beanClass);
-            if (null != path && path.isEmpty()) {
+            if (null != path && (path.isEmpty() || path.equals("*"))) {
                 rootBeans.add(bean);
             }
         }


heruan:
A different approach may be considering all mappings ending with `*`, similar to what happens with `@CDIView`.

heruan:
My first approach was an implicit wildcard mapping (e.g. considering all mapping ending with `*`), which the same with `@CDIView`. That was cleaner, but I backed to this explicit approach which cannot have side effects.

tsuoanttila:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/cdi/203#-KxqtnTwE3NSkgEGin4E:-KxqtnTwE3NSkgEGin4F:b-v89whj) (no related file):*
Hi @heruan and thanks for the PR! I'm currently taking a look at this change and considering how to proceed with it. This is indeed a problem if you want to use the PushState APIs, as they definitely do not work with CDI right now. In the FW side we now have the PushState based Navigator integrated and the way to use that is to annotate the UI with `@PushStateNavigation`. This feature kind of depends on the ability to easily find the path for the UI (something not yet implemented for the CDI) that I was planning to do using the value of the CDIUI annotation.

Would it make sense in your opinion to add a wildcard boolean to the CDIUI annotation instead of looking for a `*` in the value? This way there are less magic control characters in the path. Any ideas on the discoverability of these two approaches?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/203)*

heruan:
*[a discussion](https://reviewable.io:443/reviews/vaadin/cdi/203#-KxqtnTwE3NSkgEGin4E:-Kyzi2ibNjsV-rMCUsNY:b86m2ml) (no related file):*
<details><summary><i>Previously, tsuoanttila (Teemu Suo-Anttila) wrote…</i></summary><blockquote>

Hi @heruan and thanks for the PR! I'm currently taking a look at this change and considering how to proceed with it. This is indeed a problem if you want to use the PushState APIs, as they definitely do not work with CDI right now. In the FW side we now have the PushState based Navigator integrated and the way to use that is to annotate the UI with `@PushStateNavigation`. This feature kind of depends on the ability to easily find the path for the UI (something not yet implemented for the CDI) that I was planning to do using the value of the CDIUI annotation.

Would it make sense in your opinion to add a wildcard boolean to the CDIUI annotation instead of looking for a `*` in the value? This way there are less magic control characters in the path. Any ideas on the discoverability of these two approaches?
</blockquote></details>

Indeed I issued this before the introduction of `@PushStateNavigation`, which has been a great addition but I never had a chance to test it with CDI. If you think there's a better way to deal with this concern now, feel free to close this and proceed that way! Otherwise, I'd commit necessary changes to this to reflect recent FW releases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/203)*

Darsstar:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/cdi/203#-KxqtnTwE3NSkgEGin4E:-Kyz3-iRvpFdzXsoALff:b-2nimt5) (no related file):*
<details><summary><i>Previously, heruan (Giovanni Lovato) wrote…</i></summary><blockquote>

Indeed I issued this before the introduction of `@PushStateNavigation`, which has been a great addition but I never had a chance to test it with CDI. If you think there's a better way to deal with this concern now, feel free to close this and proceed that way! Otherwise, I'd commit necessary changes to this to reflect recent FW releases.
</blockquote></details>

I feel like the (Vaadin Spring add-on) approach of explicit wildcards (ends with '/*') to the existing URI mapping string in combination with CDI specialization is better than adding a boolean to `@CDIUI`.

As of https://github.com/vaadin/cdi/commit/924de93cc9b3f9943ff92629fc94a0be4cbb365e `CDIUIProvider` gets injected.
It could be refactored so that an `@Alternative` subclass can be activated by the end user.
Which means the current behavior could be the default for backward compatibility reasons and provide an alternative that handles `@CDIUI("/*")`, `@CDIUI("foo/*")` and `@CDIUI("/bar/baz/*")` for example.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/203)*

tsuoanttila:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/vaadin/cdi/203#-KxqtnTwE3NSkgEGin4E:-KzOcTYN94AMOSJkqk8q:b2mebab) (no related file):*
<details><summary><i>Previously, Darsstar (Dos Moonen) wrote…</i></summary><blockquote>

I feel like the (Vaadin Spring add-on) approach of explicit wildcards (ends with '/*') to the existing URI mapping string in combination with CDI specialization is better than adding a boolean to `@CDIUI`.

As of https://github.com/vaadin/cdi/commit/924de93cc9b3f9943ff92629fc94a0be4cbb365e `CDIUIProvider` gets injected.
It could be refactored so that an `@Alternative` subclass can be activated by the end user.
Which means the current behavior could be the default for backward compatibility reasons and provide an alternative that handles `@CDIUI("/*")`, `@CDIUI("foo/*")` and `@CDIUI("/bar/baz/*")` for example.
</blockquote></details>

Hi @darssta, if you take a look at the Pull Request I made on the topic, I came to the exact same conclusion and focused more on fixing the issue around the UI path mapping and providing easy to use functionality for the `@PushStateNavigation`so the user does not need to know the `/*` magic mark. 

Regarding the topic of wildcarding I made a corresponding change to Spring plug-in and noticed that actually in case of Spring you have to use `/**` to make it really the kind of wildcard we need for the feature. Both now should have more or less the same support with the annotation after the changes get reviewed and tested well enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/203)*

tsuoanttila:
#227 was merged with this feature included. Closing the PR

PR #209, opened 2017-08-20
AltCDI modifications as a single pull request

kumm:
This pull request contains all relevant modifications of AltCDI addon affects production and test code.
I've left out modifications required only to fork the add-on. Like some pom modifications, package rename, etc..
I've left out some reverted, or revised modifications too, to filter out some noise.

So this is a result of cherry pick, rewrite, squash, split of commits.
The history is cleaner, and does not conflicts. - But it's not easy to compare with the original history. -

Going to write some more details as review comments.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/209)


vaadin-cdi/src/test/java/com/vaadin/cdi/ViewDestroyTest.java
```diff
@@ -0,0 +1,65 @@
...
+    @Test
+    public void testViewChangeDestroysViewScope() throws Exception {
+        loadView(DestroyViewUI.VIEWSCOPED_VIEW);
+        assertViewDestroyCounts(0);
+        clickAndWait(DestroyViewUI.NAVIGATE_OTHER_BTN_ID);
+        assertViewDestroyCounts(1);
+    }
+
+    @Test
+    public void testNaviagteToSameViewDestroysViewScope() throws Exception {
```
kumm: At the time i was defined this behavior i didn't know the changes in vaadin 7.6.7, and 7.6.8.
see issues
https://github.com/vaadin/framework/issues/4112
https://github.com/vaadin/framework/issues/7737

CDIViewProvider.getView create a new contextual instance in a new context, so SharedUtil.equals(getCurrentView(), viewWithLongestName) in Navigator#navigateTo always false.

So navigating to the same view with same parameters is not a no-op, as intended. - intended? -
Issue fixed with https://github.com/vaadin/framework/issues/4112 might be broken this way. 
Needs further investigation.

kumm: ViewNavigationTest.testNavigationToSameViewCreatesNewContext affected too.

vaadin-cdi/src/test/java/com/vaadin/cdi/SessionContextTest.java
```diff
@@ -31,9 +32,34 @@ public void testUIsAccessSameSession() throws Exception {
...
+    public void testHttpSessionCloseDestroysSessionContext() throws Exception {
+        Assert.assertEquals(0, getCount(SessionUI.SessionScopedBean.DESTROY_COUNT));
+        clickAndWait(SessionUI.HTTP_INVALIDATEBTN_ID);
+        Assert.assertEquals(1, getCount(SessionUI.SessionScopedBean.DESTROY_COUNT));
+    }
+
+    @Test
+    @Ignore
+    //ignored because it's slow, and expiration should be same as session close
+    public void testHttpSessionExpirationDestroysSessionContext() throws Exception {
```
kumm: Revise whether
    remove this test,
    move to a slow test group
    speed it up somehow
    or just leave ignored

vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java
```diff
@@ -9,13 +9,11 @@
...
         CDIIntegrationWithConflictingDeploymentTest.class,
-        CDIIntegrationWithConflictingUIPathTest.class,
         CDIIntegrationWithDefaultDeploymentTest.class, RootViewAtContextRootTest.class,
         MultipleAccessIsolationTest.class, ScopedInstancesTest.class,
         InjectionTest.class, ConsistentInjectionTest.class,
         QualifiedInjectionTest.class, MultipleSessionTest.class,
         ScopedProducerTest.class, CrossInjectionTest.class, ShiroTest.class,
-        InappropriateNestedServletInDeploymentTest.class,
         InappropriateDeploymentTest.class, NonPassivatingBeanTest.class,
         UIDestroyTest.class})
```
kumm: On my repository I've deleted this IT. Not sure whether is it needed. 
There are a few tests - at least from me - not included here.

hesara:
Reviewed 36 of 69 files at r1.
Review status: 36 of 69 files reviewed at latest revision, 13 unresolved discussions.

---

*[vaadin-cdi/pom.xml, line 555 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3cUBZDiMubAcmjtZ1:-Ks3cUBZDiMubAcmjtZ2:b-6q4bnj) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/pom.xml#L555)):*
> ```XML
>         </dependency>
>         
>         <dependency>
> ```

Is this dependency for all profiles, or should it only be in some of them?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3P0nD4uNmX5nlDEQ0:-Ks3P0nD4uNmX5nlDEQ1:boqj7tj) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java#L41)):*
> ```Java
>  */
> @NormalUIScoped
> public class CDINavigator extends Navigator {
> ```

SpringNavigator ( https://github.com/vaadin/spring/blob/master/vaadin-spring/src/main/java/com/vaadin/spring/navigator/SpringNavigator.java ) has some more functionality and is using view activation events and overriding switchView(ViewChangeEvent) etc. to hook into the correct place for scope changes. Some of those changes could be brought to CDINavigator (probably as a separate, later step).

Some of the changes might not be compatible with Vaadin 7, though - probably not a problem for Vaadin CDI 3.0 (which will be for Vaadin Framework 8), but might be if you need to keep compatibility in older projects.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 2 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3ExK4yaeCvRXbhEP8:-Ks3ExK4yaeCvRXbhEP9:b6crpcp) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L2)):*
> ```Java
> /*
>  * Copyright 2000-2013 Vaadin Ltd.
> ```

(we should update the copyright years for the project - can do it as a separate step once this PR is merged)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-Ks3E_f69aHOrXgd1QOD:b-kqzraj) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*
> ```Java
> import java.lang.annotation.Annotation;
> 
> public class VaadinSessionScopedContext extends AbstractContext {
> ```

The class should probably have some javadoc and a since tag (probably with the value 3.0).

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 71 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3EURyjoS_n2-SQOlS:-Ks3EURyjoS_n2-SQOlT:b-7xg0xc) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L71)):*
> ```Java
>     }
> 
>     public static boolean guessContextIsUndeployed() {
> ```

(this could also have javadoc)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3EnLv9Jj6KeiNDKEO:-Ks3EnLv9Jj6KeiNDKEP:b-4bapt) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L78)):*
> ```Java
>     }
> 
> }
> ```

(Ideally should have UNIX style newline at the end of the file).

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/ViewContextualStorageManager.java, line 45 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3exL_l1021zECx7Ai:-Ks3exL_l1021zECx7Aj:bzbdukm) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/ViewContextualStorageManager.java#L45)):*
> ```Java
>     @Inject
>     private BeanManager beanManager;
>     private boolean duringBeforeViewChange;
> ```

Is this still needed if overriding Navigator.switchView()?
Perhaps it is, as the new view is potentially created before we are sure we want to switch to it.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java, line 46 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3jqgZcghOiRbcg4db:-Ks3jqgZcghOiRbcg4dc:b-b5l3px) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java#L46)):*
> ```Java
>         @Override
>         public void sessionDestroy(SessionDestroyEvent event) {
>             getLogger().fine("VaadinSessionScopedContext destroy");
> ```

(Not sure if this log entry is 100% accurate here - refers to the method called below rather than to what happens here.)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3ijzVn--KcJdenwVk:-Ks3ijzVn--KcJdenwVl:b-usmoy) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java#L78)):*
> ```Java
> 
>     @Override
>     protected void unlockSession(WrappedSession wrappedSession) {
> ```

I think rather than making this workaround here, we should probably integrate Artur's fix https://github.com/Artur-/vaadin/commit/ef2341e6570b31d289d8d3f8cd3d0baae6a04b97 to the framework.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java, line 13 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3k9jV_kzZ8jFBDNuP:-Ks3k9jV_kzZ8jFBDNuQ:bkwrfan) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java#L13)):*
> ```Java
>         CDIUIProviderTest.class, ConventionsTest.class, MultipleRootUIsTest.class,
>         CDIIntegrationWithConflictingDeploymentTest.class,
>         CDIIntegrationWithConflictingUIPathTest.class,
> ```

Why was this test removed from the IT?

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/InappropriateDeploymentTest.java, line 60 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3hTubndka1GPfkRtR:-Ks3hTubndka1GPfkRtS:b-g5vrn) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/InappropriateDeploymentTest.java#L60)):*
> ```Java
>     public void cdiUIWrongScopeBreaksDeployment() throws Exception {
>         deployer.deploy("cdiUIWrongScope");
>         fail("CDIUI that does not @UIScoped should not be deployable");
> ```

... that is not ...

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/uis/NoViewProviderNavigationUI.java, line 0 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3OILCV1cjnfBb4Xk6:-Ks3OILCV1cjnfBb4Xk7:b-hoh6f6) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/uis/NoViewProviderNavigationUI.java#L0)):*
Was NoViewProviderNavigationUI completely unused apart from being in generated WARs?

---

*[vaadin-cdi/src/test/resources/arquillian.xml, line 2 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3bYWFXR_Zz_8uuH5t:-Ks3bYWFXR_Zz_8uuH5u:b-759dr1) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/test/resources/arquillian.xml#L2)):*
> ```XML
> <arquillian xmlns="http://jboss.com/arquillian" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
>     xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
> ```

(Is there a newer version of the schema available?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 36 of 69 files reviewed at latest revision, 10 unresolved discussions.

---

*[vaadin-cdi/pom.xml, line 555 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3cUBZDiMubAcmjtZ1:-Ks4cy9bkLRi4tstc_wo:b-2pyy00) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/pom.xml#L555)):*

Is this dependency for all profiles, or should it only be in some of them?
</blockquote></details>

It is used only in ProxiedClusteringTest, but need to compile. So this test runs only with tomee-cluster, and wildfly-cluster profiles, but currently neither profile would compile without this dependency.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/ViewContextualStorageManager.java, line 45 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3exL_l1021zECx7Ai:-Ks4i8lFTNdAFxjLLb7f:b-wiq1ef) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/ViewContextualStorageManager.java#L45)):*

Is this still needed if overriding Navigator.switchView()?
Perhaps it is, as the new view is potentially created before we are sure we want to switch to it.
</blockquote></details>

Navigator#navigateTo calls ViewProvider#getView before the decision to navigate or not. That's why a new instance of the view in a new context always created before any event fired.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java, line 13 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3k9jV_kzZ8jFBDNuP:-Ks4kBnWSWSglb7xlb-y:b-g290pv) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java#L13)):*

Why was this test removed from the IT?
</blockquote></details>

I've merged together with InappropriateNestedServletInDeploymentTest, and renamed to InappropriateDeploymentTest.

Is this IT used?
Surefire run all tests even without it.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/uis/NoViewProviderNavigationUI.java, line  at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3OILCV1cjnfBb4Xk6:-Ks5GK0dM4zc6HWInkkD:bdirfpn) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/uis/NoViewProviderNavigationUI.java#L)):*

Was NoViewProviderNavigationUI completely unused apart from being in generated WARs?
</blockquote></details>

It was used in a test related to dependent CDI views. Since I've dropped dependent view support it became unused. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

hesara:
Reviewed 23 of 69 files at r1.
Review status: 59 of 69 files reviewed at latest revision, 13 unresolved discussions.

---

*[vaadin-cdi/pom.xml, line 555 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3cUBZDiMubAcmjtZ1:-Ks7SC18JJJsOsZ8ooOn:b-5ft2kl) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/pom.xml#L555)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

It is used only in ProxiedClusteringTest, but need to compile. So this test runs only with tomee-cluster, and wildfly-cluster profiles, but currently neither profile would compile without this dependency.
</blockquote></details>

OK - it is in test scope anyway so I guess we can keep it here.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java, line 109 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7uAc_vK2v1nP_dT2q:-Ks7uAc_vK2v1nP_dT2r:b-hjrslc) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java#L109)):*
> ```Java
>     @Override
>     protected void fireAfterViewChange(ViewChangeListener.ViewChangeEvent event) {
>         viewContextualStorageManager.applyChange();
> ```

I think this should probably happen in switchView(VCE) rather than fireAfterViewChange(VCE) so that view.enter() is run with the correct context.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIViewProvider.java, line 288 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7heOApmRfy7nhwTYS:-Ks7heOApmRfy7nhwTYT:bnvn7k0) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIViewProvider.java#L288)):*
> ```Java
>                     currentViewCreationalContext);
> 
>             beanManager.fireEvent(new VaadinViewCreationEvent(sessionId,
> ```

To check: these events were mostly for internal use, but I wonder if some applications might also have depended on some of them.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinExtension.java, line 117 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7fRpTmWlfXd7LiRHP:-Ks7fRpULQO2XfR7tsIA:b-uf55o5) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinExtension.java#L117)):*
> ```Java
>     }
> 
>     private void sessionClose(@Observes VaadinSessionDestroyEvent event) {
> ```

To check: are some of these events such that also some application developers may have used them.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/ViewContextualStorageManager.java, line 45 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3exL_l1021zECx7Ai:-Ks7U_BoMAdfJn9bdEuD:b32w0x) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/ViewContextualStorageManager.java#L45)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

Navigator#navigateTo calls ViewProvider#getView before the decision to navigate or not. That's why a new instance of the view in a new context always created before any event fired.
</blockquote></details>

OK

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java, line 13 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3k9jV_kzZ8jFBDNuP:-Ks7STmOuBl5iOhv3-KB:brmjk34) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/test/java/com/vaadin/cdi/DeploymentTestSuiteIT.java#L13)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

I've merged together with InappropriateNestedServletInDeploymentTest, and renamed to InappropriateDeploymentTest.

Is this IT used?
Surefire run all tests even without it.
</blockquote></details>

OK. It seems that you are right about this IT not being needed because Arquillian tests are being run by the surefire Maven plug-in, not by failsafe. I guess this is a relic from a different way to run the tests, and it should be ok just to remove the IT suite.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/uis/NoViewProviderNavigationUI.java, line  at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3OILCV1cjnfBb4Xk6:-Ks7U9uYjytVeqzb56-e:b32w0x) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/uis/NoViewProviderNavigationUI.java#L)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

It was used in a test related to dependent CDI views. Since I've dropped dependent view support it became unused. 
</blockquote></details>

OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

hesara:
Reviewed 8 of 69 files at r1.
Review status: 67 of 69 files reviewed at latest revision, 11 unresolved discussions.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewDestroyTest.java, line 48 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks8BRgoe5PUCJ-ZmxQU:-Ks8BRgoe5PUCJ-ZmxQV:b-ngr9el) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewDestroyTest.java#L48)):*
> ```Java
> 
>     @Test
>     public void testNaviagteToSameViewDestroysViewScope() throws Exception {
> ```

Should this be the case? Leaving a comment to get back to this. (Also, a typo in method name: Naviagte -> Navigate.)

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java, line 40 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks89xNeRJWz9RfDR3dY:-Ks89xNeRJWz9RfDR3dZ:b-e76awl) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java#L40)):*
> ```Java
> 
>     @Test
>     public void testNavigationToSameViewCreatesNewContext() throws Exception {
> ```

This is the bit of functionality that needs some discussion - if parameters do not change, we might want to keep the old context to avoid duplicate creation of the view on initial navigation etc. Will need to return to this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 67 of 69 files reviewed at latest revision, 11 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java, line 109 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7uAc_vK2v1nP_dT2q:-KsATXxSMXjaZrE6IiD6:bu77snl) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java#L109)):*

I think this should probably happen in switchView(VCE) rather than fireAfterViewChange(VCE) so that view.enter() is run with the correct context.
</blockquote></details>

It runs in the correct context already. ViewContextualStorageManager#getContextualStorage returns the opening view. 
But indeed, switchView is a better place. An exception in view.enter now would rollback the whole context change, because of the missed applyChange call.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

Darsstar:
Review status: 67 of 69 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java, line 32 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsJt5B8NtCNHOElKe8Y:-KsJt5B8NtCNHOElKe8Z:bgi5itz) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java#L32)):*
> ```Java
>  * </p>
>  */
> @NormalScope
> ```

Should't this be a passivating scope if the contextual instances get (indirectly) stored in the HTTP session. The CDI spec defines @SessionScoped as a passivating scope because the container can passivate HTTP sessions whenever it feels like. The container's validation doesn't 100% ensure that all contextual instances are serializable, but more deployment-time errors seems preferable to runtime errors.

And since every UI's ContextualStorage gets stored in a @VaadinSessionScoped instance, and every View's ContextualStorage gets stored in a @UINormalScoped instance can we just get rid of the pseudo scopes and only have passivating scopes? If we start firing @Initialized(X.class) and @Destroyed(X.class) for the scopes with a useful payload people should be able to create (and destroy) @Dependent scoped objects and store them in a @Application scoped registry instance.

Major breaking change, I know. BeforeBeanDiscovery.addScope() could maybe be used for opt-in backwards compatibility?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-KsLGEz3QK4zZFBQzUGk:bw24502) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*

The class should probably have some javadoc and a since tag (probably with the value 3.0).
</blockquote></details>

Implementing AlterableContext (CDI 1.1) would be nice since CDI 1.2 is a dependency. AbstractContext#destroy returns a boolean so that means either an custom implementation (don't know how copying code from one APv2 project to another APv2 project involves, might be worth it?) of wrapping a subclass and ignoring the return value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 67 of 69 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java, line 32 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsJt5B8NtCNHOElKe8Y:-KsPgDY_CoJjeEjxfacT:b-6snjuf) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java#L32)):*
<details><summary><i>Previously, Darsstar (Dos Moonen) wrote…</i></summary><blockquote>

Should't this be a passivating scope if the contextual instances get (indirectly) stored in the HTTP session. The CDI spec defines @SessionScoped as a passivating scope because the container can passivate HTTP sessions whenever it feels like. The container's validation doesn't 100% ensure that all contextual instances are serializable, but more deployment-time errors seems preferable to runtime errors.

And since every UI's ContextualStorage gets stored in a @VaadinSessionScoped instance, and every View's ContextualStorage gets stored in a @UINormalScoped instance can we just get rid of the pseudo scopes and only have passivating scopes? If we start firing @Initialized(X.class) and @Destroyed(X.class) for the scopes with a useful payload people should be able to create (and destroy) @Dependent scoped objects and store them in a @Application scoped registry instance.

Major breaking change, I know. BeforeBeanDiscovery.addScope() could maybe be used for opt-in backwards compatibility?
</blockquote></details>

I have some arguable reasons to avoid passivating scope:
* backward compatibility
* i see it as a feature. It is optional to keep the session serializable. I have to bother with it when it matters. 

HTTP Session, and CDI @sessionscoped is two different things. My implementation indirectly stores everything in HTTP Session, but nothing in @sessionscoped. Storing non-serializable objects in HTTP session is not a best practice, but it is not against any spec.
With disabled session persistence on application server it is acceptable, and won't lead to runtime errors. 

About pseudo scopes, i'm sure we can't get rid of them. Passivating scopes are normal scopes. It means CDI implementation proxies bean instances. 
Vaadin framework don't handle - and i think should not - proxied UIs, and components properly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 59 of 69 files reviewed at latest revision, 14 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java, line 109 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7uAc_vK2v1nP_dT2q:-KsZ6j8OgY287tOtGxms:b-leooad) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java#L109)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

It runs in the correct context already. ViewContextualStorageManager#getContextualStorage returns the opening view. 
But indeed, switchView is a better place. An exception in view.enter now would rollback the whole context change, because of the missed applyChange call.
</blockquote></details>

As you can see i've ended up using fireBeforeViewChange. Due to refactoring for beforeLeave support, switchView is too late. A point before showView need to apply context change.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-KsZCgsdQD0qswul56cY:b-23jnhq) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*
<details><summary><i>Previously, Darsstar (Dos Moonen) wrote…</i></summary><blockquote>

Implementing AlterableContext (CDI 1.1) would be nice since CDI 1.2 is a dependency. AbstractContext#destroy returns a boolean so that means either an custom implementation (don't know how copying code from one APv2 project to another APv2 project involves, might be worth it?) of wrapping a subclass and ignoring the return value.
</blockquote></details>

Other contexts are affected too. Would be nice, but is there a use-case that worth it to wrap, or copy deltaspike?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 71 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3EURyjoS_n2-SQOlS:-KsZPSLYNftoMvpoIc91:b-fq38ps) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L71)):*

(this could also have javadoc)
</blockquote></details>

introduced some javadoc

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3EnLv9Jj6KeiNDKEO:-KsZPOZhkljEqAiIn-IB:b-7ql1td) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L78)):*

(Ideally should have UNIX style newline at the end of the file).
</blockquote></details>

fixed

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java, line 46 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3jqgZcghOiRbcg4db:-KsZRZunpr5DuOGKdJGe:b-5c3wew) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java#L46)):*

(Not sure if this log entry is 100% accurate here - refers to the method called below rather than to what happens here.)
</blockquote></details>

moved to where it happens

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/InappropriateDeploymentTest.java, line 60 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3hTubndka1GPfkRtR:-KsZPoTMNJBfKCvA8B02:b-7ql1td) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/InappropriateDeploymentTest.java#L60)):*

... that is not ...
</blockquote></details>

fixed

---

*[vaadin-cdi/src/test/resources/arquillian.xml, line 2 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3bYWFXR_Zz_8uuH5t:-KsZOMLjnMFiEAQnazqQ:b-j70o9z) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/test/resources/arquillian.xml#L2)):*

(Is there a newer version of the schema available?)
</blockquote></details>

did not find, but namespace was wrong ( or old )

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

Darsstar:
@hesara Just curious, what kind of backward compatibility breaking changes would 3.0.0 allow?
* An upgrade to CDI 2.0? That would sort of mirror dropping Java 7.
* The session not serializing in some situations when it did before?
* Switching to a different approach to parsing the URI, but providing the old method as a disabled alternative bean?
* Switching from UriFragment base navigation to history api based navigation by default but providing the URIFragment one as a disabled alternative bean?
* other stuff you like to see changed?
  

---

Review status: 59 of 69 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java, line 32 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsJt5B8NtCNHOElKe8Y:-KsXQwStNq5rJjdXfZMr:b-8b007k) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java#L32)):*
First of all, I am 100% for storing contextual instances in the session.

>backward compatibility

Having them remain pseudo scopes is bad form. Injecting a non-serializable `@UIScoped`/`@NormalUIScoped`/`@ViewScoped`/`@NormalViewScoped` instance into an `@ApplicationScoped` class, something that causes no problems in 2.0.0, results in the session serialization failure. So backward compatibility is broken anyway. 
Non-serializable `@UIScoped`/`@NormalUIScoped`/`@ViewScoped`/`@NormalViewScoped`observer methods being created because of an event would also prevent serialization. 

>i see it as a feature. It is optional to keep the session serializable. I have to bother with it when it matters.

Then remove the `@Scope` and `@NormalScope` annotations and start using BeforeBeanDiscovery .addScope() instead. The boolean values would default to true, but should be based on properties, if present. That way the end user can opt-in for pseudo scopes. CDI support in IDEs won't be as useful though.

Or I guess we could wrap all relevant `Bean`s, `BeanAttributes` or `AnnotationType`s, not sure which, using the relevant events and override getScope(), or replace scope annotations in the case of `AnnotatedType`, to return a different scope when the user decides to opt-out of passivation and/or normal-scope. Then they can just ignore the IDE warnings knowing that at runtime the bean will have a different scope.

The add-on could provide two portable extensions the end user needs to activate.

>HTTP Session, and CDI @sessionscoped is two different things. My implementation indirectly stores everything in HTTP Session, but nothing in @sessionscoped.

Since Weld and OpenWebBeans store `@SessionScoped` instances in the HTTP Session they two very related things. Due to the UIContextualStorageManager being `@VaadinSessionScoped` the ContextualStorage for UI scoped instances is effectively `@VaadinSessionScoped` too. Which I should have made more clear, sorry.

>About pseudo scopes, i'm sure we can't get rid of them. Passivating scopes are normal scopes. It means CDI implementation proxies bean instances.
Vaadin framework don't handle - and i think should not - proxied UIs, and components properly.

CDI implementations also have to create proxies when there are interceptors or decorators associated with the managed bean.
The problems with proxies I can think of are hashCode() and equals() implementations , not having a non-private no-args constructor and synchronization blocks due to proxies. 

There are already hashCode() and equals() implementations in the Vaadin framework that keep CDI in mind.Why not more?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209#-:-KsZm4uVHs7DdquoQM2J:b-37zx44)*

kumm:
Review status: 59 of 69 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java, line 32 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsJt5B8NtCNHOElKe8Y:-KseCgUM2ox2inTqU9ou:b28px49) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java#L32)):*
>  Injecting a non-serializable @UIScoped/@NormalUIScoped/@ViewScoped/@NormalViewScoped instance into an @ApplicationScoped class, something that causes no problems in 2.0.0, results in the session serialization failure

Why should it cause serialization error in an @ApplicationScoped class? Do you mean injecting into a @SessionScoped class?
Yes non serializable @UIScoped, and @ViewScoped cause session serialization failure, because injected by reference. Exactly like in 2.0.0. 
But @NormalUIScoped, @NormalViewScoped serialization simply does not happen in 2.0.0, so you lost the bean states after deserialization without any error. Now it happens, so it cause session serialization failure.  

Was broken silently - is broken loudly.  Won't call it a compatiblity break.
With passivating scope your good old app won't deploy, it is a compatibility problem. Though i've introduced some, so just one more problem.

> the end user can opt-in for pseudo scopes

Do not. It should be mandatory for UIs, and vaadin components. See below.

> CDI implementations also have to create proxies when there are interceptors or decorators

Should be true, but for pseudo scope the proxy have to hold a direct reference to the bean instance. This way the main problem with proxies does not exist. Would need some test whether are there any problem.

Client Proxies for normal scoped beans are problematic because the framework don't know which instance will be called. The call is forwarded by the proxy to the bean instance for the active context.

Not imaginary examples:

1. A for loop in the core framework iterating through all uis in the vaadin session and calling a setter/getter on the ui. What will happen with proxies of normal scoped uis?
a) somehow there is an active ui context. All calls will be forwarded to the same bean instance in the active context.
b) no active context. ContextNotActiveException

2. The framework does some housekeeping on the closed UI.
Calling through the proxy will get ContextNotActiveException, the Context already closed with the UI.


Workarounds in 2.0.0 for these, and similar problems were the root cause of the bugs and inconsistencies i've started to fix. 
After a lot of trial and error, i've decided to throw workarounds away, and do not support normal scoped uis. Simply does not worth it.

Did not search similar problems with components, but sure warnings in 2.0.0 for normal scoped components are not without any reason.

I don't see hashCode() and equals() implementations can solve these kind of problems.
By the way normal scoped CDI proxy won't delegate equals and hashCode to the bean instance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

Darsstar:
Review status: 59 of 69 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java, line 32 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsJt5B8NtCNHOElKe8Y:-KshiTS-2KaBn5I7sBuE:bsfa09) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java#L32)):*
>Why should it cause serialization error in an @applicationscoped class? Do you mean injecting into a @sessionscoped class?

No, I meant `@ApplicationScoped` since those contextual instances don't get stored in the session in both 2.0.0 and this branch. In 2.0.0 the injected VaadinCDI scoped bean instance would never get serialized because the context it is stored in never gets serialized. In this branch the instance gets stored in a context which might get serialized.

>Client Proxies for normal scoped beans are problematic because the framework don't know which instance will be called. The call is forwarded by the proxy to the bean instance for the active context.

Should have realized this myself... I did not... Now I feel stupid :p

>Workarounds in 2.0.0 for these, and similar problems were the root cause of the bugs and inconsistencies i've started to fix.
After a lot of trial and error, i've decided to throw workarounds away, and do not support normal scoped uis. Simply does not worth it.

Point taken. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 59 of 69 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java, line 32 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsJt5B8NtCNHOElKe8Y:-KsjZVoQ48630hUGCALS:bohkypr) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinSessionScoped.java#L32)):*
> No, I meant @ApplicationScoped since those contextual instances don't get stored in the session in both 2.0.0 and this branch. In 2.0.0 the injected VaadinCDI scoped bean instance would never get serialized because the context it is stored in never gets serialized. In this branch the instance gets stored in a context which might get serialized.

Right, injecting non serializable normal UI/View scoped beans to anywhere in the application will cause session serialization error. What i don't see is what was better in 2.0.0.

HTTP sessions don't "might get" serialized. 
Looked at java servlet spec. Serialization mentioned only for distributed environments.

Look it by practice. It is controlled by application server configuration.
a) turned off - no serialization will occur
b) session reload on redeploy enabled -  serialization occures at application undeploy/ app server shutdown
c) session replication enabled - session serialization occures at the end of every request

a) this branch: no problem, 2.0.0: no problem
b) this branch: generate error in the log, sessions lost, users are sad.
b) 2.0.0: no error, sessions serialized, and reloaded. 
But your applications state is inconsistent. You have your vaadin ui deserialized with the active view, but you lost all your normal scoped beans for that view context. 
You will get error reports about insane things, users are sad, you are sad too.
c) this branch: generate error in the log, application keep working, session distribution fails, on a failover user sessions from this node are lost.
c) 2.0.0 no error, session replication happens.
But your application state is invalid on other nodes. After failover same error reports as b) .

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

hesara:
Reviewed 8 of 10 files at r2.
Review status: 67 of 69 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIViewProvider.java, line 288 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7heOApmRfy7nhwTYS:-Ksmj9ME5CbljVRyqUBI:beem8mh) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIViewProvider.java#L288)):*

To check: these events were mostly for internal use, but I wonder if some applications might also have depended on some of them.
</blockquote></details>

Based on discussion elsewhere, we should probably introduce view activation events like those in the Spring add-on, and can remove the current internal events as long as the release notes make this clear. If there is demand for them, we can later introduce more lifecycle events etc.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinExtension.java, line 117 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks7fRpTmWlfXd7LiRHP:-KsmjchrjE_zGLSjHpGP:b-bv730s) ([raw file](https://github.com/vaadin/cdi/blob/3d0bca72217c3b0d0a39000bf7524643276825a1/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinExtension.java#L117)):*

To check: are some of these events such that also some application developers may have used them.
</blockquote></details>

See the comment in CDIViewProvider - we should be able to drop the internal events, and maybe add new public and documented ones later if needed.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-Ksmqz3mhI9ypniqbC0M:bkt11z8) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

Other contexts are affected too. Would be nice, but is there a use-case that worth it to wrap, or copy deltaspike?
</blockquote></details>

AlterableContext would be nice as that would make Vaadin CDI more of a "first class citizen" in the CDI world. 

CDI spec method AlterableContext.destroy() does not return anything (CDI 1.1 - 2.0), whereas a DeltaSpike implementation in DeltaSpike AbstractContext inherited here does, so is this the conflict that might require copying code?
Is there no compatible version of DeltaSpike available?
Would a trivial wrapper context class eliminate the need for copying?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3ijzVn--KcJdenwVk:-Ksmk91U6u-WICIfBoic:b-suls7s) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java#L78)):*

I think rather than making this workaround here, we should probably integrate Artur's fix https://github.com/Artur-/vaadin/commit/ef2341e6570b31d289d8d3f8cd3d0baae6a04b97 to the framework.
</blockquote></details>

in https://github.com/vaadin/framework/pull/9851 (pending a test fix).

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java, line 101 at r2](https://reviewable.io:443/reviews/vaadin/cdi/209#-KsmqVz7gE-eV36b3hWQ:-KsmqVz7gE-eV36b3hWR:bsykm7q) ([raw file](https://github.com/vaadin/cdi/blob/cf1b89f930372ac17b84b259d5e46e1ef1763d38/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java#L101)):*
> ```Java
>         // navigation is delayed in DelayNavigationView.beforeLeave()
>         clickAndWait(ViewNavigationUI.SUCCESS_NAV_BTN_ID);
>         // target view is contructed
> ```

(constructed)

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/uis/ViewNavigationUI.java, line 208 at r2](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ksmpao7Dfc5bWqJLJqD:-Ksmpao7Dfc5bWqJLJqE:b-oo3u5t) ([raw file](https://github.com/vaadin/cdi/blob/cf1b89f930372ac17b84b259d5e46e1ef1763d38/vaadin-cdi/src/test/java/com/vaadin/cdi/uis/ViewNavigationUI.java#L208)):*
> ```Java
> 
>         @Override
>         public void beforeLeave(ViewBeforeLeaveEvent event) {
> ```

This feels borderline bending the rules as the navigation based on the event is pushed far into the future and the user can do other things in between, but probably ok for the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

Darsstar:
Review status: 67 of 69 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-Kshyp7FYL9SfyiUN_7u:buema8b) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*

AlterableContext would be nice as that would make Vaadin CDI more of a "first class citizen" in the CDI world. 

CDI spec method AlterableContext.destroy() does not return anything (CDI 1.1 - 2.0), whereas a DeltaSpike implementation in DeltaSpike AbstractContext inherited here does, so is this the conflict that might require copying code?
Is there no compatible version of DeltaSpike available?
Would a trivial wrapper context class eliminate the need for copying?

</blockquote></details>

[https://github.com/apache/deltaspike/blob/2d4e4099852ffbd6aec1aca73dda331cf5329d91/deltaspike/core/api/src/main/java/org/apache/deltaspike/core/util/context/AbstractContext.java](DeltaSpike 1.8.0), which is the most recent, only has an incompatible version.

The spec also encourages custom extension to fire `@Initialized(X.class)`, `@Destroyed(X.class)` as of 1.1 and as of 2.0 `@BeforeDestroyed(X.class)` when a custom context is initialized, destroyed and about to be destroyed, respectively.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 67 of 70 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-KsohrShdxp2hUBxS7a1:b-w8uspp) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*
> Would a trivial wrapper context class eliminate the need for copying?

Good question. There was already a wrapper. So trivial, i've pushed it.

> The spec also encourages custom extension to fire `@Initialized(X.class)`, `@Destroyed(X.class)` as of 1.1

I like the idea to fire lifecycle events. With some payload they can be useful. I'am thinking about if we can use `@Destroyed(ViewScoped.class)` instead of the planned view deactivation event. View activation event still needed.

>  and as of 2.0 `@BeforeDestroyed(X.class)`

I think it is too early to depend on CDI 2.0.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

kumm:
Review status: 66 of 70 files reviewed at latest revision, 7 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-Ksqu8QlFsPbEkY7KzZm:b-3g86xj) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*
> There was already a wrapper. So trivial, i've pushed it.

And of course this context is the only one not extending the wrapper :)
But extending it seems trivial too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

hesara:
I'd vote for merging this PR to the target branch vaadin/cdi:kumm-alt-cdi (possibly after addressing some of the minor issues still open as comments in Reviewable here), and then making additional PRs to that branch for further changes. We could also make GitHub issues for the points that might still need to be addressed (e.g. porting additional features from Vaadin Spring to CDINavigator, release notes commenting on incompatibilities). Otherwise, we risk the discussion in this big review stretching for a long time.

@darsstar @kumm any objections or things you'd still like changed before the initial merge? We can also still change things merged from here after the merge if needed, and this would also make it easier for @darsstar to make PRs to the branch.

---

Reviewed 2 of 2 files at r3.
Review status: 68 of 70 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-Ksr3DyVo-By0utbskaH:b-cwtnns) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

> There was already a wrapper. So trivial, i've pushed it.

And of course this context is the only one not extending the wrapper :)
But extending it seems trivial too.

</blockquote></details>

It seems VaadinExtension already wraps all the other custom contexts but not this one. Otherwise looks ok to me with the wrapper.

The lifecycle events would be nice, but can be added in a separate later PR as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209#-:-Ksr3ZwR8I0gX3CB26QP:bnfg0a4)*

hesara:
As for the discussion on equals() and hashCode() with proxies, I'm not sure if everything was already clear, but just in case: those are a real problem for Vaadin components - partly also because some CDI implementations have (intentionally) broken versions of those for proxies. This means that e.g. only a proxy of a component can be added as a child of another component, but the component will later attempt to remove itself from its parent (using an un-proxied reference). I can't recall where such things happen, but they definitely did happen.

kumm:
> any objections or things you'd still like changed before the initial merge?

I've just pushed to wrap vaadin session context, to finish what i've begun.
It is OK to me to merge.

---

Review status: 67 of 70 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209#-:-KsrBSUAKJcVr_CEcYq9:bx950fz)*

Darsstar:
No objections.

---

Review status: 67 of 70 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209#-:-KsrVEGyhkaT739afDeI:b-oewjzq)*

hesara:
Reviewed 2 of 10 files at r2, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java, line 41 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3P0nD4uNmX5nlDEQ0:-Ksresn5Nxrg--YR3RyV:b54fvms) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/CDINavigator.java#L41)):*

SpringNavigator ( https://github.com/vaadin/spring/blob/master/vaadin-spring/src/main/java/com/vaadin/spring/navigator/SpringNavigator.java ) has some more functionality and is using view activation events and overriding switchView(ViewChangeEvent) etc. to hook into the correct place for scope changes. Some of those changes could be brought to CDINavigator (probably as a separate, later step).

Some of the changes might not be compatible with Vaadin 7, though - probably not a problem for Vaadin CDI 3.0 (which will be for Vaadin Framework 8), but might be if you need to keep compatibility in older projects.
</blockquote></details>

To address after merge as a separate PR.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java, line 30 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3E_f69aHOrXgd1QOC:-KsrgPGXaL6XEDQC2o6G:b-s5tcm4) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/internal/VaadinSessionScopedContext.java#L30)):*

It seems VaadinExtension already wraps all the other custom contexts but not this one. Otherwise looks ok to me with the wrapper.

The lifecycle events would be nice, but can be added in a separate later PR as well.
</blockquote></details>

To address in a later PR

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks3ijzVn--KcJdenwVk:-Ksrfo1AwqrTTF4RpPgv:b-23mptb) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/main/java/com/vaadin/cdi/server/VaadinCDIServletService.java#L78)):*

in https://github.com/vaadin/framework/pull/9851 (pending a test fix).
</blockquote></details>

To remove in a later PR after the framework patch is merged (if it makes it to a suitable version).

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewDestroyTest.java, line 48 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks8BRgoe5PUCJ-ZmxQU:-Ksrg6HzKhjT-4RJ2afV:b9l3bm8) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewDestroyTest.java#L48)):*

Should this be the case? Leaving a comment to get back to this. (Also, a typo in method name: Naviagte -> Navigate.)
</blockquote></details>

to address later separately

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java, line 40 at r1](https://reviewable.io:443/reviews/vaadin/cdi/209#-Ks89xNeRJWz9RfDR3dY:-KsrgK6TTdx9l1ScM96Z:b-5henfh) ([raw file](https://github.com/vaadin/cdi/blob/ec7a4945bac8021d152fcb6152696a7883daa1b6/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java#L40)):*

This is the bit of functionality that needs some discussion - if parameters do not change, we might want to keep the old context to avoid duplicate creation of the view on initial navigation etc. Will need to return to this.
</blockquote></details>

to decide on later

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/209)*

PR #211, opened 2017-09-10
Some CDIView annotation cleanup

kumm:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/211)

PR #212, opened 2017-09-10
View context strategies

kumm:
A feature to control the lifespan of view scoped context.
Short description:
`@CDIView(value = "myviewA", contextStrategy = ViewName.class)`
The view context opened by this view behaves like view contexts in the current CDI addon (2.0). Remains open while the same view is active regardless of parameters.
`@CDIView(value = "myviewB", contextStrategy = ViewNameAndParameters.class)`
The view context opened by this view behaves in par with view navigation in vaadin framework. Remains open while the same view is active with same parameters. This is the default.
`@CDIView(value = "myviewC", contextStrategy = Always.class)`
The view context opened by this view remains open until the next navigation event. Regardless of view name and parameters. This is the current behaviour.

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/cdi/212)


hesara:
Partial review - I will review the rest of the implementation, and ask for API feedback also from someone else. This should also be on top of the changes merged from #211.

---

Reviewed 8 of 19 files at r1.
Review status: 8 of 19 files reviewed at latest revision, 5 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 82 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtknMPZ_PmRcP4AATQz:-KtknMPZ_PmRcP4AATR-:bgoo8q0) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L82)):*
> ```Java
>      * @return a strategy class
>      */
>     public Class<? extends ViewContextStrategy> contextStrategy() default ViewContextStrategy.ViewNameAndParameters.class;
> ```

Which should the default strategy be? Is this equivalent to the framework default behavior or something else?

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/AbstractViewStrategyTest.java, line 99 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtklpveDyxqzGjTGkjb:-KtklpveDyxqzGjTGkjc:bl1xaai) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/AbstractViewStrategyTest.java#L99)):*
> ```Java
>     }
> 
>     protected void assertToOtherViewContextCreated(String sourceState, String srcBeanValue)
> ```

What does "To" mean here? Should it be clarified by making the method name a bit longer and more explicit?

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/AbstractViewStrategyTest.java, line 115 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktklxfdc53fQEGQCO1A:-Ktklxfdc53fQEGQCO1B:bl1xaai) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/AbstractViewStrategyTest.java#L115)):*
> ```Java
>     }
> 
>     protected void assertToTestedViewContextCreated(
> ```

What does "To" mean here? Should it be clarified by making the method name a bit longer and more explicit?

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java, line 63 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktkkj4u-NcHJkZCmdqo:-Ktkkj4u-NcHJkZCmdqp:b-a5e6hx) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewNavigationTest.java#L63)):*
> ```Java
> 
>     @Test
>     public void testAfterViewChangeCDIFiredInNewContext() throws Exception {
> ```

the method name is not fully clear to me (maybe something like testAfterViewChangeCDI*Event*FiredInNewContext() would be clearer)

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java, line 43 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtklLSdWyvoXHtaPEMW:-KtklLSdWyvoXHtaPEMX:b5pmx0x) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java#L43)):*
> ```Java
> 
>     @Test
>     public void testViewNameStrategyUpAfterUiInit() throws Exception {
> ```

What does "Up" in this method name and the one below mean?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212#-:-KtkowCHHUGQnr0Xac-i:b-5ajnhy)*

hesara:
Reviewed 8 of 19 files at r1.
Review status: 16 of 19 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 37 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktkv3chIPoDVXCF4H1U:-Ktkv3chIPoDVXCF4H1V:b-5umbqg) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L37)):*
> ```Java
> 
>     /**
>      * Whether active context contains target navigation state.
> ```

This and/or the class level javadoc could perhaps explain a bit more what "contains" means. (not a blocker, reasonably clear)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 54 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuweleM9ZdkUSHW51:-KtkuweleM9ZdkUSHW52:b-17tnug) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L54)):*
> ```Java
>      * - {@link com.vaadin.navigator.View#enter(ViewChangeEvent)} will be called again
>      * on the same view instance.
>      * - Navigator view change events does not mean a view context change.
> ```

maybe "... do not..." as in plural

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 72 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuSF30UzTdmgEYFGz:-KtkuSF30UzTdmgEYFH-:b-9kg2ab) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L72)):*
> ```Java
>     /**
>      * Strategy to hold the context open while
>      * view name and view parameters does not change.
> ```

"does" -> "do" (plural)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 76 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuYERSZzr6SY_uWV-:-KtkuYERSZzr6SY_uWV0:b-yd04vy) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L76)):*
> ```Java
>      * This strategy is on par with navigator view lifecycle.
>      * - After all {@link ViewChangeEvent#beforeViewChange(ViewChangeEvent)}
>      * is called - if navigation is not reverted -, new view context is activated.
> ```

"is" -> "are" if in plural ("all"); this sentence is rather complicated and could be easier to read if structured a bit differently

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuodolnLhPC3tBIpL:-KtkuodpzWbjAXvh_ILd:b-1y3py9) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L78)):*
> ```Java
>      * is called - if navigation is not reverted -, new view context is activated.
>      * - After all {@link ViewChangeEvent#afterViewChange(ViewChangeEvent)}
>      * is called, old view context is closed.
> ```

see comment on line above

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 110 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkqnfuMYO7iwlZLP2T:-KtkqnfuMYO7iwlZLP2U:b-v021wc) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L110)):*
> ```Java
>      * even when parameters does not change.
>      */
>     @RequestScoped
> ```

Could this scope cause problems if it can be used in a background thread and an access block? Is it possible for this to be called so?

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java, line 29 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktku3QF4HXFE4EqmWvl:-Ktku3QF4HXFE4EqmWvm:b-uw5cat) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java#L29)):*
> ```Java
> import org.junit.Test;
> 
> public class ViewStrategiesUiInitTest extends AbstractManagedCDIIntegrationTest {
> ```

(this only tests the positive cases, not that the ViewNameStrategies ever return false - probably ok as long as covered by other tests)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

kumm:
Review status: 2 of 19 files reviewed at latest revision, 12 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 82 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtknMPZ_PmRcP4AATQz:-KtmbusLrf2Ajku2n71i:b-7m6vev) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L82)):*

Which should the default strategy be? Is this equivalent to the framework default behavior or something else?
</blockquote></details>

For the second question, yes it is equivalent to framework default behaviour ( as far as i understand ).

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuodolnLhPC3tBIpL:-KtmaOkyWyTvVh63Kjcb:b-99r4qn) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L78)):*

see comment on line above
</blockquote></details>

Sorry for my english. I will retry it.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 110 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkqnfuMYO7iwlZLP2T:-Ktm_mRKhz2PEFqaUr3Y:b-ee4xpj) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L110)):*

Could this scope cause problems if it can be used in a background thread and an access block? Is it possible for this to be called so?
</blockquote></details>

You are right, it can cause problems. Changing to @NormalUIScoped.

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java, line 29 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktku3QF4HXFE4EqmWvl:-KtmY7dXLP_g6dbBpAj4:b-i9sps6) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java#L29)):*

(this only tests the positive cases, not that the ViewNameStrategies ever return false - probably ok as long as covered by other tests)
</blockquote></details>

Basically it is a smoke test. Maybe i should rename to ViewStrategiesSmokeTest.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

Legioth:
Review status: 2 of 19 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-KtojBe6FiIjrtIjl8Cz:b-yvza4) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
> ```Java
>      * @return a strategy class
>      */
>     public Class<? extends ViewContextStrategy> contextStrategy() default ViewContextStrategy.Always.class;
> ```

I'm slightly concerned with the discoverability of this API approach: how can a user know that there are built-in implementations of the class and where to find them?

I would instead suggest an `enum` with values for all the built-in options.

If there is reason to expect that developers would also want to provide their own strategy implementations, then we could also introduce a `contextStrategyClass` parameter in the annotation and throw deploy-time exceptions if both are defined.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

kumm:
Review status: 2 of 19 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-KtqZXcMePc3ZZ4W6YPh:bfb7kmd) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

I'm slightly concerned with the discoverability of this API approach: how can a user know that there are built-in implementations of the class and where to find them?

I would instead suggest an `enum` with values for all the built-in options.

If there is reason to expect that developers would also want to provide their own strategy implementations, then we could also introduce a `contextStrategyClass` parameter in the annotation and throw deploy-time exceptions if both are defined.
</blockquote></details>

Indeed, built-in implemetations are almost hidden.
Maybe some little changes can help (to some degree).

1. move implemetations to a class named ViewContextStrategies as static inner classes
2. refer this class from javadoc

Does it address your concerns?
If not, of course we can go with the additional enum parameter and deploy time check.

I think own strategies can be useful. For example one can extend the scope of the view context beyond a single view. Something like a conversation scope.
So i would keep it open. And when built-in implemetations are classes, it is for free.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

Legioth:
Review status: 2 of 19 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-Ktu-utQ4reyflPnYJQ0:b-jvlg9k) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

Indeed, built-in implemetations are almost hidden.
Maybe some little changes can help (to some degree).

1. move implemetations to a class named ViewContextStrategies as static inner classes
2. refer this class from javadoc

Does it address your concerns?
If not, of course we can go with the additional enum parameter and deploy time check.

I think own strategies can be useful. For example one can extend the scope of the view context beyond a single view. Something like a conversation scope.
So i would keep it open. And when built-in implemetations are classes, it is for free.


</blockquote></details>

It's always easier to find the values of a specific enum than to find subclasses of some given class, even though the changes you suggest would certainly help.

What about renaming this property to `contextStrategyClass`? We can then keep an eye open for usability issues and if we want to introduce an enum to make it easier to find the built-in strategies, then we can add `contextStrategy` without breaking backwards compatibility.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

kumm:
Review status: 2 of 19 files reviewed at latest revision, 6 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-Ktu0GA0TpfTEi8gai3Q:bcwu87b) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, Legioth (Leif Åstrand) wrote…</i></summary><blockquote>

It's always easier to find the values of a specific enum than to find subclasses of some given class, even though the changes you suggest would certainly help.

What about renaming this property to `contextStrategyClass`? We can then keep an eye open for usability issues and if we want to introduce an enum to make it easier to find the built-in strategies, then we can add `contextStrategy` without breaking backwards compatibility.
</blockquote></details>

I've thinked about custom implementation classes. I can let it go.
Keep it simple. We can add it later, when really need. ( even me would not use to make a conversion scope )

It is good to me to go with a single enum. It makes the parameter foolproof.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

hesara:
Reviewed 2 of 15 files at r2, 5 of 5 files at r3, 5 of 12 files at r4, 6 of 6 files at r5.
Review status: 16 of 19 files reviewed at latest revision, 9 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/AfterViewChange.java, line 30 at r6](https://reviewable.io:443/reviews/vaadin/cdi/212#-Kto__D9liwabsDrdrZx:-Kto__D9liwabsDrdrZy:b-oq3ei5) ([raw file](https://github.com/vaadin/cdi/blob/048390b93b257d8fd18d5b45a5ca27ea8e81586b/vaadin-cdi/src/main/java/com/vaadin/cdi/AfterViewChange.java#L30)):*
> ```Java
> 
> /**
>  * {@link ViewChangeListener.ViewChangeEvent} can be observed with this qualifier.
> ```

(To check: does javadoc render links etc. correctly on the first line/paragraph? I have the recollection the first line is often used as a plain text short description in some contexts, with no formatting applied.)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 78 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtoSyd1MiEj6nC4RRWD:-KtoSyd1MiEj6nC4RRWE:b-sors50) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L78)):*
> ```Java
> 
>     /**
>      * The strategy to decide whether to open a new context while the
> ```

to discuss: should this be linked to the current view so it can decide what to do also when the active view changes

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-KtzEKbXasfAPlvqyzXG:blfv52g) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

I've thinked about custom implementation classes. I can let it go.
Keep it simple. We can add it later, when really need. ( even me would not use to make a conversion scope )

It is good to me to go with a single enum. It makes the parameter foolproof.
</blockquote></details>

We discussed this a little with @legioth .
The academically correct solution would probably be separate annotations for each of the modes, which would also allow extensibility in the future without breaking backwards compatibility. Even though an enum might provide better discoverability, separate annotations for each of the cases (used together with the CDIView annotation) would be a more natural way in CDI environments. How does this sound to you?

---

*[vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java, line 29 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktku3QF4HXFE4EqmWvl:-KtoTo7BamjBvfJ5SOzy:b-5mw8vi) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/test/java/com/vaadin/cdi/ViewStrategiesUiInitTest.java#L29)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

Basically it is a smoke test. Maybe i should rename to ViewStrategiesSmokeTest.
</blockquote></details>

I guess it is ok like this as well - the method name clarifications also help.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

kumm:
Review status: 16 of 19 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-Ku-s8pXcbmmbeuw8gBn:b-a91o40) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*

We discussed this a little with @legioth .
The academically correct solution would probably be separate annotations for each of the modes, which would also allow extensibility in the future without breaking backwards compatibility. Even though an enum might provide better discoverability, separate annotations for each of the cases (used together with the CDIView annotation) would be a more natural way in CDI environments. How does this sound to you?
</blockquote></details>

Let's see it in code.
We have a view something like this:
`
@CDIView
@ViewNameDriven
class MyVew implements View {...}`

To know what annotation defines a view context strategy we need another annotation, to annotate it. So we have annotations like these:

`public @interface ViewContextStrategyParsed {}`

`@ViewContextStrategyParsed public @interface ViewNameDriven {}`

This way we can search for an annotation annotated by ViewContextStrategyParsed on the view class.

And we have a strategy class. Have to annotate it too. Would be nice to annotate with a CDI qualifier parametrized by the annotation class on the view. This way CDI lookup would be simple. But it would mean one more annotation. 
It is simpler to annotate with same annotation, but this way we will need some code to discover strategy classes.
`  @NormalUIScoped
    @ViewNameDriven
    class ViewName implements ViewContextStrategy {...}
`

Indeed, the API for the end user getting nicer.
The implementation will gain some complexity, and we lost one thing, the default strategy is not visible from the code.

But i like syntactic sugar, so i like it. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

hesara:
Please rebase (or cherry-pick) this change on top of latest master.

kumm:
Review status: 2 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-KuGSeAsVKqj4vXcjk52:bc35kmq) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

Let's see it in code.
We have a view something like this:
`
@CDIView
@ViewNameDriven
class MyVew implements View {...}`

To know what annotation defines a view context strategy we need another annotation, to annotate it. So we have annotations like these:

`public @interface ViewContextStrategyParsed {}`

`@ViewContextStrategyParsed public @interface ViewNameDriven {}`

This way we can search for an annotation annotated by ViewContextStrategyParsed on the view class.

And we have a strategy class. Have to annotate it too. Would be nice to annotate with a CDI qualifier parametrized by the annotation class on the view. This way CDI lookup would be simple. But it would mean one more annotation. 
It is simpler to annotate with same annotation, but this way we will need some code to discover strategy classes.
`  @NormalUIScoped
    @ViewNameDriven
    class ViewName implements ViewContextStrategy {...}
`

Indeed, the API for the end user getting nicer.
The implementation will gain some complexity, and we lost one thing, the default strategy is not visible from the code.

But i like syntactic sugar, so i like it. 
</blockquote></details>

I have an implementation. Push it here, or as a separate PR later?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

kumm:
Review status: 2 of 19 files reviewed at latest revision, 8 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-KuuWuVri6EZC7yBtCYn:b-t5vshr) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

I have an implementation. Push it here, or as a separate PR later?
</blockquote></details>

You can take a look at the last two commit [here](https://github.com/wbstr/vaadin-alt-cdi/commits/strategy-annotations). 
There are some javadoc fixes too according to your comments.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

hesara:
Marked some more changes (after rebase) as reviewed. Some old comments are still open.

---

Reviewed 3 of 5 files at r8, 5 of 12 files at r9, 6 of 6 files at r10, 1 of 1 files at r11.
Review status: 17 of 19 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212#-:-Kw-pG2A1QTDOJLC2lFe:b-o3cwo2)*

tsuoanttila:
I have some updates to the JavaDocs in my workspace that I'd want to push to be part of this review. Can you enable maintainers to push into your branch so I can commit them? (Should be a checkbox in the PR page)

kumm:
Checked "Allow edits from maintainers". It should work now.

tsuoanttila:
@kumm how busy are you? If we want to get this to 3.0 we'll need to do the annotation based naming thing as well. How soon do you think you'd have the time to rebase and make a PR if we merge this really soon?

kumm:
@tsuoanttila: It would be nice to get in 3.0. 
So i'm gong to rebase and create a pull request of (this)[https://github.com/wbstr/vaadin-alt-cdi/commits/strategy-annotations]. 
Today, after business hours.

tsuoanttila:
We will most likely be unable to merge this yet, but if you can open up the next PR on top of this branch, I can rebase it on top of master after this one gets merged given you check the allow maintainers thingie for the new PR as well!

pleku:
Reviewed 2 of 19 files at r1, 2 of 2 files at r12.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 29 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-KzxfvA2fGA-CrhrN8Pu:-KzxfvA2fGA-CrhrN8Pv:b4u1pou) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L29)):*
> strategy whether

_strategy on whether_

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 31 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-Kzxg7LBDMv1ccn_KK6z:-Kzxg7LBDMv1ccn_KK7-:brbhmwq) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L31)):*
> a new one created

_a new one is created_ (not 100% sure)

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 33 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-KzxwcMt1syBewxb02TR:-KzxwcMt1syBewxb02TS:bvu1z2t) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L33)):*
> Implementations of this interface need a scope since they are instantiated by CDI.

_Implementations of this interface need to be scoped since they are instantiated by the CDI container._

What happens if you leave it out ? Is there a default scope ? What scope should I use ?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 39 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-KzxwxpXpm-86wn52zym:-KzxwxpXpm-86wn52zyn:b-ijpu6v) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L39)):*
> ```Java
> 
>     /**
>      * Returns whether active context contains target navigation state. This
> ```

the active context

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 60 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-KzxxcJs6omXv2lbu3HE:-KzxxcJs6omXv2lbu3HF:b-zclw9v) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L60)):*
> events do not mean a view context change.

_events don't necessarily cause a view context change_ ? 
_events do not mean that the view context will change_ ?

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 115 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-Kzxxx-rTAkINiNmdU6O:-Kzxxx-rTAkINiNmdU6P:b-fn985b) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L115)):*
>  release the context and create a new one.

_releases the context and creates a new one._

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

kumm:
Ok, i see it.
Since there are more doc changes in the queue, and a new discussion is opened too, i think it is better to rebase, and send a new PR a little later.

---

Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 33 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-KzxwcMt1syBewxb02TR:-KzzLLQGNduMeSiYfMpm:b3ise68) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L33)):*
>What happens if you leave it out ? Is there a default scope ? 

Yes, there is. The default scope is dependent in CDI.

Strategy class instantiated simply by Deltaspike [BeanProvider::getContextualReference](https://deltaspike.apache.org/javadoc/1.7.2/org/apache/deltaspike/core/api/provider/BeanProvider.html#getContextualReference-java.lang.Class-java.lang.annotation.Annotation...-). 
You can see the reasons in the doc. 

We would have to define the life-cycle of dependent beans, and implement it.
Currently dependent beans are working, but a warning is issued in developement, and unitTest Deltaspike project stages. Bean instance is created before the view is instantiated. I know one pitfall, PreDestroy never called because it is not destroyed properly - but does not cause a leak -.

>What scope should I use ?

Basically it depends on the state you store in the bean. 

But, if you use UI.acces within background threads you are responsible for activating the desired context. UI.access itself activates automatically - indirectly, by setting the ThreadLocals - only the vaadin contexts . Core framework use UI.access too in a request of an other UI - though i think only for cleanup -. Non vaadin scopes are dangerous.

View scoped beans are meaningless, they are created in the view context before the view change.

So we have three safe scopes VaadinSessionScoped, NormalUIScoped, UIScoped.
Because there is no reason to use non-normal (non proxied) scopes i would left out UIScoped.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212#-:-KzzavlndkRKJs_q4N7d:bpnhiba)*

tsuoanttila:
Reviewed 2 of 19 files at r1.
Review status: 18 of 19 files reviewed at latest revision, 10 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 54 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuweleM9ZdkUSHW51:-Kwj-7HSFnmynZxbqbLk:b-896fix) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L54)):*

maybe "... do not..." as in plural
</blockquote></details>

Done.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 72 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuSF30UzTdmgEYFGz:-KwiywsNCJ8ltp9wIOJh:b-896fix) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L72)):*

"does" -> "do" (plural)
</blockquote></details>

Done.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 76 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuYERSZzr6SY_uWV-:-KwiyyKm-TvH0j5IbvuP:b-896fix) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L76)):*

"is" -> "are" if in plural ("all"); this sentence is rather complicated and could be easier to read if structured a bit differently
</blockquote></details>

Done.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 78 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtkuodolnLhPC3tBIpL:-Kwiyz2Bg87b0jN3jkSN:b-896fix) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L78)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

Sorry for my english. I will retry it.
</blockquote></details>

Done.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 29 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-KzxfvA2fGA-CrhrN8Pu:-L-2-gQa0uOQruJdpanj:bb6un1j) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L29)):*
<details><summary><i>Previously, pleku (Pekka Hyvönen) wrote…</i></summary><blockquote>

> strategy whether

_strategy on whether_
</blockquote></details>

Done

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 31 at r12](https://reviewable.io:443/reviews/vaadin/cdi/212#-Kzxg7LBDMv1ccn_KK6z:-L-2-l1z3exQUvDQ46l8:bagr5z4) ([raw file](https://github.com/vaadin/cdi/blob/80a867a6cf88d801b7c3b4b0a99e2000d76e95e4/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L31)):*
<details><summary><i>Previously, pleku (Pekka Hyvönen) wrote…</i></summary><blockquote>

> a new one created

_a new one is created_ (not 100% sure)
</blockquote></details>

Seems a bit better.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

pleku:
Reviewed 1 of 1 files at r13.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 37 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktkv3chIPoDVXCF4H1U:-L-6flZzAX5Qch_RNXDi:bnc9v3l) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L37)):*

This and/or the class level javadoc could perhaps explain a bit more what "contains" means. (not a blocker, reasonably clear)
</blockquote></details>

@tsuoanttila is this clarified on the class level ? if so, should probably add `@see` linking to it ? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

tsuoanttila:
*[vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java, line 37 at r1](https://reviewable.io:443/reviews/vaadin/cdi/212#-Ktkv3chIPoDVXCF4H1U:-L-BaD-A02HhMNfvKO_F:b-xzj4hg) ([raw file](https://github.com/vaadin/cdi/blob/f26725d977dad9126de58da8549cbf17e09a3e86/vaadin-cdi/src/main/java/com/vaadin/cdi/ViewContextStrategy.java#L37)):*
<details><summary><i>Previously, pleku (Pekka Hyvönen) wrote…</i></summary><blockquote>

@tsuoanttila is this clarified on the class level ? if so, should probably add `@see` linking to it ? 
</blockquote></details>

I tried to explain it here, the "contains" is a tricky word.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

pleku:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-L-CUF6XobOaP_YtNHV2:b-c1oz) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, kumm wrote…</i></summary><blockquote>

You can take a look at the last two commit [here](https://github.com/wbstr/vaadin-alt-cdi/commits/strategy-annotations). 
There are some javadoc fixes too according to your comments.

</blockquote></details>

@tsuoanttila it seems that these two commits from @kumm are not incorporated into this PR ?
https://github.com/wbstr/vaadin-alt-cdi/commits/strategy-annotations

Should we just merge this PR and create another PR with those commits ? They seem to be API breaking so it's either now or not for 3.0 (or never ?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

tsuoanttila:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-L-CbYVfY2SYREi4Ubfp:b-q8frkb) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, pleku (Pekka Hyvönen) wrote…</i></summary><blockquote>

@tsuoanttila it seems that these two commits from @kumm are not incorporated into this PR ?
https://github.com/wbstr/vaadin-alt-cdi/commits/strategy-annotations

Should we just merge this PR and create another PR with those commits ? They seem to be API breaking so it's either now or not for 3.0 (or never ?)
</blockquote></details>

I'd say close all discussions, merge as is, make the follow up PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

pleku:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java, line 83 at r2](https://reviewable.io:443/reviews/vaadin/cdi/212#-KtojBe6FiIjrtIjl8Cy:-L-D80UZzIP2ua7m4rw2:bvzhb4b) ([raw file](https://github.com/vaadin/cdi/blob/d08a70f2c5d6286509f7dc323cb3a1f8cfa3c599/vaadin-cdi/src/main/java/com/vaadin/cdi/CDIView.java#L83)):*
<details><summary><i>Previously, tsuoanttila (Teemu Suo-Anttila) wrote…</i></summary><blockquote>

I'd say close all discussions, merge as is, make the follow up PR.
</blockquote></details>

Allright, please do the honors

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/cdi/212)*

tsuoanttila:
@kumm it's merged, waiting for the annotation patch!

kumm:
I've commented about a javadoc change [here](https://github.com/vaadin/cdi/commit/e8dc32e913705b3b5d93c1779f863ea31f70ff5f#r26015196).
Going to send a pull request too without the mentioned documentation change. I can reabase if you change it.

Issue #324, opened 2020-03-30
UIScoped RouterLayout has issue on re-entering


roeltje25:
Hi,

I have a Routerlayout which is @UIScoped and it contains a @RouteScoped component with the RouterLayout as its parent layout via @Route annotation. In this nested component I trigger a navigation to another route which does not contain this parent layout. Afterwards I click the back button in the browser and the child component is added to the RouterLayout for a second time. 

When inspeciting calls to showRouterLayout and removeRouterLayout. I see that the @UIScoped routerlayout is removed per removeRouterLayout and then the new compoennt is added with showRouterLayout. which is ok. Then I click back and I see the new component gets removed, and the old @Uiscoped layout is added back (remember that the nested child was never removed) then the system adds a new copy of the @Routescoped child which results in double element in the page.

I hope this makes sense, please let me know if more information is required.
Roel

roeltje25:
So please find attached the minimal example of the issue. you should be able to run it using mvn tomee:run
The issue talked about here is demonstrated by clicking on the text of oneview and then clicking on anotherview, and repeat this a few times. You will see oneview text gets repeated. Which should not happen if I understand the javadoc of RouterLayout showRouterLayoutContent correctly.

Additional issue is that the root page (connected to OneView via RouterAlias("")) does not seem to generate the required ParentLayout at all (which I would expect it to). Just found this out while testing the minimal example.
[parentlayout-issue-demo.zip](https://github.com/vaadin/cdi/files/4456421/parentlayout-issue-demo.zip)

roeltje25:
Additionally, although I have this in a CDI project. I think this may not be a CDI issue, but an issue in the method com.vaadin.flow.component.internal.UIInternals.showRouteTarget (line 652 in UIInternals.java).  when you switch from a @route component ONE with parentlayout A nested withing parentlayout B to @route component TWO directly with parentlayout B. and then you go back, this code tries to remove @Route component ONE from parentlayout B, while it should actually remove parentlayout A frorm parentlayout B. At least, this is my hunch after a quick investigation.

roeltje25:
*bump*

roeltje25:
still *bump*

denis-anisimov:
First of all:
in the example `@RouteScoped` annotation is used without any `RouteScopeOwner`  annotation which makes  `@RouteScoped`  no-op .
So I indeed doubt that this issue is a CDI issue.

[...]

PR #344, opened 2020-06-11
remove out-of-scope components from parent

roeltje25:
When HasElements / Components are out of scope they are destroyed, but are not removed from their parents. In case of UIScoped parent layouts with RouteScoped router content for example, this will result in these elements remaining in the component tree with undesirable results. This change checks if objects that will become out of scope and should be destroyed are Component or HasElement and will remove them from their parent to ensure they can ultimately be garbage collected.

This pull request is associated to issue #345 and related to issue #324

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/context/AbstractContextualStorageManager.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/context/AbstractContextualStorageManager.java
@@ -22,6 +22,7 @@ import org.apache.deltaspike.core.util.context.ContextualStorage;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.HasElement;
+import com.vaadin.flow.dom.Element;
 import com.vaadin.flow.router.RouterLayout;
 
 import javax.annotation.PreDestroy;
@@ -111,7 +112,9 @@ abstract class AbstractContextualStorageManager<K> implements Serializable  {
             if( maybeRouterLayout.isPresent( ) )
                 return;
         }
-        willBeRemoved.getElement( ).removeFromParent( );
+        Element element = willBeRemoved.getElement( );
+        if(element != null)
+            element.removeFromParent( );
     }
 
     protected Set<K> getKeySet() {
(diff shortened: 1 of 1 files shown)


roeltje25:
am creating associated issue at this moment

roeltje25:
Denis, thank you for your comments. ALthough, reading your comments it seems you are refering to my proposed patch in #324. This pullrequest is actually a different angle to the problem and only handles removing Out-of-scope components from their parent. Something which seems to me to be strange that is is not the case already. But we discussed this in #324 to some point already.

denis-anisimov:
Well, yes. I kept in mind the issue with router.

But does this PR has any sense at all? 
How the component may be destroyed if it's directly referenced from another bean (regardless of any UI consideration).

roeltje25:
Without this PR the RouteScoped component is removed from the scope, but it will remain attached to the UIScoped parent layout. This PR will make sure it can get destroyed by making it remove it from the parent first. Attaching to the parent layout is handled by the framework, so developers do not explicitly think about removing it after the child component is out of scope. This will make the child component get removed when it is out of scope. which to me and my colleagues clearly makes a lot of sense. How does it seem non-sensical to you? Perhaps we can ask another set of eyes to have a look. It seems to me that perhaps I am not explaining myself correctly.

roeltje25:
This has been running in production in our servers for a month now and running exactly as we expect it to. Is there a possiblity for another review?

It seems ownership of attaching and detaching components for managed beans like @RouteScoped and @UIScoped beans should be with the framework not the developer of the application, that would go against the point of "managed beans". As it stands @RouteScoped components are not removed from @UISCoped parent layouts when the @RouteScoped bean goes out of scope by moving to another route. It should as the documentation of @RouteScoped mentions that in the absence of @RouteScopeOwner annotations the route scope owner is the bean itself (which is out of scope after moving to another route).

Looking forward to a review....

mshabarov:
Thanks for contribution, @roeltje25 !
The issue has been fixed on the [Flow side](https://github.com/vaadin/flow/pull/10973) based on your [suggestion](https://github.com/vaadin/cdi/issues/324#issuecomment-641778167).
Thus, I close this PR.

roeltje25:
You are very welcome. And I am very happy this has been solved.

Op ma 17 mei 2021 13:50 schreef Mikhail Shabarov ***@***.***>:

PR #359, opened 2020-12-23
refactor: fix unit tests after app-config changes

pleku:
Adds necessary mocking for application configuration for CdiServlet
in unit tests.

Related to vaadin/flow#9642

### DIFF
--- a/vaadin-cdi/src/test/java/com/vaadin/cdi/CdiVaadinServletTest.java
+++ b/vaadin-cdi/src/test/java/com/vaadin/cdi/CdiVaadinServletTest.java
@@ -23,13 +23,18 @@ import javax.servlet.ServletContext;
 import javax.servlet.ServletException;
 
 import java.util.Collections;
+import java.util.Enumeration;
 
+import com.vaadin.flow.server.startup.ApplicationConfiguration;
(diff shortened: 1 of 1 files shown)


vaadin-cdi/src/test/java/com/vaadin/cdi/CdiVaadinServletTest.java
```diff
@@ -61,6 +66,14 @@ public void setUp() throws ServletException {
         Mockito.when(servletContext.getAttribute(Lookup.class.getName()))
                 .thenReturn(lookup);
 
+        final DefaultApplicationConfigurationFactory applicationConfigurationFactory
```
denis-anisimov, line 69: There is no need to mock `DefaultApplicationConfigurationFactory` or its interface.
It's enough to properly mock `ServletContext`:
the `ApplicationConfiguration`  is retrieved from the context by the attribute `ApplicationConfiguration.class.getName()`.

If there is no attribute value then the factory is called. But if the attribute is given right away then it will be just returned without asking the factory.

But this is not really important. It just allows to reduce a number of lines.

denis-anisimov (approved):

PR #376, opened 2021-09-02
fix: implement getActiveContextualStorages properly for RouteScope co…

### LINKED ISSUE #375: RouteScopedContext inherits getActiveContextualStorages which needs to be reimplemented for it since it works wrong
`RouteScopedContext::getActiveContextualStorages` is inherited from `AbstractContext`.
The implementation in the `AbstractContext` passes a `null` as a Contextual/first arg into `getContextualStorage` method.
It works for many scopes but it doesn't work for `RouteScopedContext` : the passed `Contextual` is in fact used in the logic of the method (unlike impls for the other scopes) and `null` is unexpected there.
Moreover in the `RouteScopedContext` getActiveContextualStorages may really returns   a number  `ContextualStorage`s (which are active) whereas the other scopes impls returns only one active `ContextualStorage` instance.

There is no unit test for `getActiveContextualStorages` for `RouteScopedContext` that's why the issue is not revealed.

Here is the implementation which should be used for `RouteScopedContext`:

```
@Override
    protected List<ContextualStorage> getActiveContextualStorages() {
        return getStorageManager().getActiveContextualStorages();
    }

```

And `ContextualStorageManager` (nested class) should be extended with method: 
```
private List<ContextualStorage> getActiveContextualStorages() {
            return getKeySet().stream().filter(
                    key -> key.getUIId().equals(getUIStoreId(UI.getCurrent())))
                    .map(key -> getContextualStorage(key, false))
                    .collect(Collectors.toList());
        }
```

And unit test should be added to test `getActiveContextualStorages`/ `destroyAllActive`.

denis-anisimov:
…ntext

fixes #375

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
@@ -151,6 +151,13 @@ public class RouteScopedContext extends AbstractContext {
                     new RouteStorageKey(key.getOwner(), getUIStoreId(ui)));
         }
 
+        private List<ContextualStorage> getActiveContextualStorages() {
+            return getKeySet().stream().filter(
+                    key -> key.getUIId().equals(getUIStoreId(UI.getCurrent())))
+                    .map(key -> getContextualStorage(key, false))
+                    .collect(Collectors.toList());
+        }
+
         private String getUIStoreId(UI ui) {
             ExtendedClientDetails details = ui.getInternals()
                     .getExtendedClientDetails();
@@ -247,6 +254,11 @@ public class RouteScopedContext extends AbstractContext {
     @Override
     public boolean isActive() {
         return isUIContextActive.get();
+    }
+
+    @Override
+    protected List<ContextualStorage> getActiveContextualStorages() {
+        return contextManager.getActiveContextualStorages();
     }
 
     @Override
--- a/vaadin-cdi/src/test/java/com/vaadin/cdi/context/RouteContextualStorageManagerTest.java
+++ b/vaadin-cdi/src/test/java/com/vaadin/cdi/context/RouteContextualStorageManagerTest.java
@@ -18,11 +18,14 @@ package com.vaadin.cdi.context;
 
 import javax.annotation.PreDestroy;
 import javax.enterprise.event.Event;
+import javax.enterprise.inject.spi.BeanManager;
 import javax.inject.Inject;
 import javax.inject.Provider;
 
 import java.util.Collections;
+import java.util.List;
 
+import org.apache.deltaspike.core.util.context.ContextualStorage;
 import org.apache.deltaspike.testcontrol.api.junit.CdiTestRunner;
 import org.junit.After;
 import org.junit.Assert;
@@ -115,6 +118,9 @@ public class RouteContextualStorageManagerTest {
     private LocationChangeEvent changeEvent;
 
     private NavigationData data = Mockito.mock(NavigationData.class);
+
+    @Inject
+    private BeanManager manager;
 
     @Before
     public void setUp() {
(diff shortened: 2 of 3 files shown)


vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
```diff
@@ -151,6 +151,13 @@ private void relocate(UI ui, RouteStorageKey key) {
                     new RouteStorageKey(key.getOwner(), getUIStoreId(ui)));
         }
 
+        private List<ContextualStorage> getActiveContextualStorages() {
+            return getKeySet().stream().filter(
+                    key -> key.getUIId().equals(getUIStoreId(UI.getCurrent())))
+                    .map(key -> getContextualStorage(key, false))
```
mshabarov, line 157: Shouldn't this method also check that the route's owner is in the current navigation chain?

vaadin-cdi/src/test/java/com/vaadin/cdi/context/RouteContextualStorageManagerTest.java
```diff
@@ -243,6 +249,38 @@ public void onBeforeEnter_anotherUIHasNoWindowName_beanIsDestroyedOnUiDestroy()
...
+
+        Mockito.when(event.getNavigationTarget())
+                .thenReturn((Class) Group1.class);
+        beforeNavigationTrigger.fire(event);
+
+        MemberOfGroup1 bean1 = memberOfGroup1.get();
+
+        storages = context.getActiveContextualStorages();
+        Assert.assertEquals(1, storages.size());
+        // Now the second UI is active and its storage is a different one
```
mshabarov, line 280: I don't get the meaning of prerequisites for this assertion.
`UI ui = doSetUp("foo", null);` creates a new UI with corresponding window name  "foo", adds it to the session and sets it as active.
`doSetUp(null, ui.getSession());` creates another UI with the same window name as previous and sets it as active, the previous one is not active anymore.
Then you check that the active storage being returned by context corresponds to the second UI.
Is that right?

vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
denis-anisimov: route's owner of WHAT ?
route's owner is something that defines a bean . There is no any bean in the context.
The method name is `getActiveContextualStorages` which mean : get all active storages. And there is no beans, injection points , etc.
The `UI` contains an active navigation chain. Based on this chain all the storages which has a key associated with the active `UI` instance should be returned.
I don't even understand your question.

vaadin-cdi/src/test/java/com/vaadin/cdi/context/RouteContextualStorageManagerTest.java
denis-anisimov: `doSetUp(null, ui.getSession())` creates _another_ UI with not defined window name: because it's not specified and mock which returns window name is not called.

The main purpose here is : create TWO different UIs.
`doSetUp()` activates the last created UI: it makes it active via set `UI:setCurrent`. But this is an impl details.
`doSetUp` makes active the `UI` which it creates. As a result there are two `UI`s and only one is active.
And test checks exactly what's implemented in the code: that only storages of the second UI are returned when method `getActiveContextualStorages` is called.
The storage of the first `UI` ( which is inactive) is not returned.
So prerequisites are : two different UIs.

vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
mshabarov: `ContextualStorage`s of route scope are associated with corresponding `RouteStorageKey` which includes the owner of the bean and the UI id this bean belongs to (or, more precisely, it's lifespan belong to). That's why I asked about the owner, owner of a bean. But I believe `getActiveContextualStorages` is called in the codes which don't care about the owner of a bean (I found only one place - in `destroyAllActive`), and, thus, you don't use it.
Anyway, thanks for explanations.

mshabarov (approved):

PR #383, opened 2022-01-14
Remove deltaspike dependency

manolo:
It's only used in couple of tests.


manolo:
Cannot remove the dependency

PR #392, opened 2022-08-11
fix: Add @Stereotype to scopes

### LINKED ISSUE #303: @RouteScoped beans are not discovered when bean-discovery-mode="annotated"
When using bean-discovery-mode="annotated" all beans with Pseudo-scopes are not discovered or considered as beans. In order to mitigate this RouteScoped would have to be a Stereotype for example. This is e.g. discussed in https://issues.jboss.org/browse/WELD-1733?_sscc=t

```java
@Stereotype
@Scope
@Inherited
@Target({ ElementType.ANNOTATION_TYPE, ElementType.TYPE, ElementType.FIELD,
        ElementType.METHOD, ElementType.CONSTRUCTOR })
@Retention(RetentionPolicy.RUNTIME)
public @interface RouteScoped {
}
```

### LINKED ISSUE #391: Constructor Injection only works with CDI Beans
According to the documenation at https://vaadin.com/docs/latest/integrations/cdi/instantiated-beans
> If the type is not found as a CDI bean – for example, because it is ambiguous or does not have a no-arguments public constructor – instantiation falls back to the default Vaadin behavior, and the component is instantiated as a plain-old Java object (POJO). Dependency injection is performed after instantiation, and injects still work, but other CDI features do not. The reason for this is that the instantiated component is not a contextual instance.

Actually, if the type is not found as a CDI bean
- a no-args or var-args constructor is required, see [ReflectTools.createInstance(Class)](https://github.com/vaadin/flow/blob/master/flow-server/src/main/java/com/vaadin/flow/internal/ReflectTools.java#L480)
- only field injection is performed, see [AbstractCdiInstantiator.getOrCreate(Class)](https://github.com/vaadin/cdi/blob/master/vaadin-cdi/src/main/java/com/vaadin/cdi/AbstractCdiInstantiator.java#L67)

Consequently

- constructor injection only works with CDI beans
- the flow-cdi-tutorial and the examples on https://vaadin.com/docs/latest/integrations/cdi/contexts only work with [bean-discovery-mode="all"](https://github.com/vaadin/flow-cdi-tutorial/blob/master/src/main/webapp/WEB-INF/beans.xml) because they use constructor injection

Please update the documentation or treat it as a bug.

TatuLund:
Fixes #391
Fixes #303


mcollovati (commented):
Two observations:
- according to the CDI specs normal scope are bean defining annotations, so no need for the Stereotype annotation
- I'm uncertain if it is correct to mark a scope with Stereotype. Specs say that a stereotype annotations may have a scope, but I cannot find an indication or example to make a scope a stereotype.

Second one is just a thought, I need to further investigate on it

TatuLund:
@mcollovati you seem to be correct. The normal scopes do not need it. I made a test build and seems to work. There is no detrimental effect if the annotation is there however. I do not have strong opinions regarding that one.

mcollovati:
@TatuLund my opinion is that if we want beans to be discovered they should use a normal scope  (e.g. `NormalRouteScoped`)
Other solution would be *defining* a specific stereotype annotation with the required default scope 

If we take a look at the link in the related issue, we can see that the suggested way is to define a stereotype annotation, not to mark a scope as stereotype

```java
@javax.enterprise.inject.Stereotype
@javax.inject.Singleton
@Target(TYPE)
@Retention(RUNTIME)
public @interface CDISingleton {

}
```

If there's the need for a bean with pseudo-scope (e.g `RouteScoped`) to be automatically discovered, then we need to implement a specific stereotype annotation and use it or maybe have a producer method.

But this is just my opinion, I may be wrong, refusing the scope to be a stereotype.
So, I would like to see some additional comments on this.
Ping @mshabarov, @caalador, @taefi and @MarcinVaadin

TatuLund:
@mcollovati you may be correct or this is opionated how this should be done.

In Vaadin 8 this was done indirectly a bit in a way you suggest. There is `@CDIView` annotation, which has `@ViewScoped` and adds `@Stereotype`

https://vaadin.com/start/v8/full-stack-spring

> we can see that the suggested way is to define a stereotype annotation

So the `@CDIView` is defacto such annotation (also `@UIScope`)

It should in theory be possible that vaadin-cdi would have `CDIRouter` extending `Router`, which adds similar annotation, say `CDIRoute`. But that is substantially bigger change.

roeltje25:
> Two observations:
> 
> * according to the CDI specs normal scope are bean defining annotations, so no need for the Stereotype annotation
> * I'm uncertain if it is correct to mark a scope with Stereotype. Specs say that a stereotype annotations may have a scope, but I cannot find an indication or example to make a scope a stereotype.
> 
> Second one is just a thought, I need to further investigate on it

Note that @RouteScoped is not a normal scope and is as such not a bean defining annotation according to the spec.

Not having this annotation bea bean defining annotation is a bit strange though? That's the whole point of the annotation. I want it to be injectable under the route scope....

I vote for the RouteScoped annotation to be a stereotype.

mcollovati:
> Note that @RouteScoped is not a normal scope and is as such not a bean defining annotation according to the spec.

Yes, sure. My comment was about the PR changes, adding Stereotype to all scope annotation, including normal scoped. I could have been clearer, sorry

roeltje25:
> Yes, sure. My comment was about the PR changes, adding Stereotype to all scope annotation, including normal scoped. I could have been clearer, sorry

Then we are in agreement

taefi:
Actually, I'm waiting for the reply tp this comment here:
https://github.com/vaadin/cdi/issues/303#issuecomment-1223884419

mstahv:
@kumm, would you have a moment to check this change and give your opinion on it? We are currently missing a bit of CDI expertise in the team.

kumm:
It does not seem to be the intended use of `@Stereotype`, however a smart approach to make pseudo scopes bean defining. Unfortunately I've never tried it, and hard to decide if it is safe across all cdi implementations.
Maybe it would worth it taking a try implementing in the extension.
Per spec normal scopes should not need any modifications.

taefi:
Thanks, @TatuLund for the contribution. In the end, we decided not to annotate `@RouteScoped` and `@UIScoped` with `@Stereotype`, and instead, we added `@CdiComponent`, so I propose to close this PR in favor of #397.

roeltje25:
Well, although I would have given quite some pushback if I were in the decision making process.... I can live with this approach.  Thanks for picking up the original issue #303 and coming up with a way forward

mshabarov:
Closed in favour of https://github.com/vaadin/cdi/pull/397

PR #393, opened 2022-08-11
fix: throw on not existing scope only if bean creation is requested

### LINKED ISSUE #390: Conditional observers in @RouteScoped components not working with Weld?
We are using Vaadin 23 (23.0.9) with vaadin-cdi (14.0.0) and are having problems with conditional observers in @RouteScoped components.
We are running on Tomcat 9 and use Weld (3.1.9.Final) as our CDI implementation. Our current problem is that we get the following exception when firing an event that is observed in a @RouteScoped component:

> java.lang.IllegalStateException: Route owner 'class XXX' instance is not available in the active navigation components chain: the scope defined by the bean 'YYY' doesn't exist.

We believe that the problem are the changes implemented in Vaadin CDI 13.0.0 (https://github.com/vaadin/cdi/releases/tag/13.0.0):

> Trying to use @RouteScope when there is no active route component present is not allowed.

In the Weld implementation we are using, the method getReceiverIfExists (in class ObserverMethodImpl) calls the method getContextualStorage() in class RouteScopedContext. And getReceiverIfExists is called before the reception of the observer is considered. Therefore all classes that have an observer on the event are trying to be received and since not all @RouteScoped classes are in the current navigation chain, the above mentioned error is thrown.

```
Object receiver = getReceiverIfExists(null); //this is where the exception is thrown
if (receiver == null && reception != Reception.IF_EXISTS) {
   // creational context is created only if we need it for obtaining receiver
   // ObserverInvocationStrategy takes care of creating CC for parameters, if needed
   creationalContext = beanManager.createCreationalContext(declaringBean);
   receiver = getReceiverIfExists(creationalContext);
}
if (receiver != null) {
   sendEvent(event, receiver, creationalContext);
}
```
Right now, we are not able to use conditional observers in @RouteScoped classes. Is there a way to fix this?
Link to example project: https://github.com/nluckas/vaadincdiwithweldexample

mcollovati:
## Description

If bean creation is not requested, just return a null value from context storage
instead of throwing an exception because route owner instance is not available
in current navigation components chain

Fixes #390

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [X] I have added tests to ensure my change is effective and works as intended.

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
@@ -27,6 +27,7 @@ import java.lang.annotation.Annotation;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -264,11 +265,6 @@ public class RouteScopedContext extends AbstractContext {
     @Override
     protected ContextualStorage getContextualStorage(Contextual<?> contextual,
                                                      boolean createIfNotExist) {
-        RouteStorageKey key = convertToKey(contextual, createIfNotExist);
-        return contextManager.getContextualStorage(key, createIfNotExist);
-    }
-
-    private RouteStorageKey convertToKey(Contextual<?> contextual, boolean createIfNotExist) {
         Bean<?> bean = getBean(contextual);
         UI ui = UI.getCurrent();
         Class<?> owner = getOwner(ui, bean);
@@ -278,7 +274,8 @@ public class RouteScopedContext extends AbstractContext {
                             + "active navigation components chain: the scope defined by the bean '%s' doesn't exist.",
                     owner, bean.getBeanClass().getName()));
         }
-        return contextManager.getKey(ui, owner);
+        RouteStorageKey key = contextManager.getKey(ui, owner);
+        return contextManager.getContextualStorage(key, createIfNotExist);
     }
 
     private boolean navigationChainHasOwner(UI ui, Class<?> owner) {
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
@@ -93,7 +93,7 @@ public class RouteScopedContext extends AbstractContext {
         }
 
         private void destroyDescopedBeans(UI ui,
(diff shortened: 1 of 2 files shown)


roeltje25 (changes requested):
Thank you very mich @mcollovati  I need this change as well. I am currently working around it using Bytebuddy. I think however that considering separation of concern that the convertKey method should not throw with an IllegalStateException at all. Although there is a possible problem in fetching the ContextualStorage for a particular key, the key itself can be generated.

I suggest to move the validity check to the  `getContextualStorage()` method. That way we do not pollute `convertToKey` with a parameter that does not affect the key generation.

vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
```diff
@@ -263,16 +263,16 @@ protected List<ContextualStorage> getActiveContextualStorages() {
...
     }
 
-    private RouteStorageKey convertToKey(Contextual<?> contextual) {
+    private RouteStorageKey convertToKey(Contextual<?> contextual, boolean createIfNotExist) {
         Bean<?> bean = getBean(contextual);
         UI ui = UI.getCurrent();
         Class<?> owner = getOwner(ui, bean);
-        if (!navigationChainHasOwner(ui, owner)) {
+        if (!navigationChainHasOwner(ui, owner) && createIfNotExist) {
             throw new IllegalStateException(String.format(
```
roeltje25, line 276: So pertaining to my earlier comment: `return null` here and throw in `getContexstualStorage` after line 267 if `createIfNotExist=true` and `key=null`

mcollovati: Thanks for suggestion.
Moving the check to the `getContexstualStorage` method means removing `convertToKey` at all, since to fill the exception message, we need all the code above.
Convert key will have only `contextManager.getKey(ui, owner)` and it doesn't make sense to have a method with this single line

roeltje25:
Fair enough. It will do fine like this.

Op vr 12 aug. 2022 07:00 schreef Marco Collovati ***@***.***>:

roeltje25 (approved):
With the inlining of convertToKey it looks perfect to me. Hopefully it will get merged soon.

mshabarov (approved):

mshabarov:
Will be released together with https://github.com/vaadin/cdi/pull/392, as soon as we conclude the best way to do that.

PR #397, opened 2022-09-07
feat: enable bean discovery when bean-discovery-mode="annotated"

### LINKED ISSUE #303: @RouteScoped beans are not discovered when bean-discovery-mode="annotated"

### LINKED ISSUE #391: Constructor Injection only works with CDI Beans

taefi:
## Description
@CdiComponent added to be used on types
that are supposed to be discovered and
managed by CDI container when
bean-discovery-mode is set to "annotated"
in beans.xml

Fixes #303
Fixes #391

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] I have performed self-review and corrected misspellings.

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.


roeltje25:
Thanks for providing a solution to #303 this way., although I would have argued fervently for #392. But reading the specs I may interpret the intentions for @Stereotype a bit broader than you guys do. Fair enough.  Thanks for picking this up!

roeltje25:
Ok, one more little argument for why #392 is actually a good solution and is not a case of not-intended use of @Stereotype. Please look at the implementation of e.g. the @Decorator stereotype in package javax.decorator. This is a stereotype, in order to be found by CDI, but it does not declare any additional scope, @Alternative, or @Named annotation. Clearly, the  meaning of the annotation is to make @Decorator a bean defining annotation. 
Also a @Stereotype can according to its javadoc have 0 interceptor bindings, 0 scope annotations, 0 alternative and 0 named and 0 stereotypes.

Ok. anyway. I will stop.

vaadin-cdi-itest/src/test/java/com/vaadin/cdi/itest/BeanDiscoveryModeTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @OperateOnDeployment("bean-discovery-mode-all")
+    public void beanDiscoveryWorks_when_beanDiscoveryModeIsAll() {
+        validateBeanDiscoveryAndInjectionWorks();
+    }
+
+    private void validateBeanDiscoveryAndInjectionWorks() {
+        click(SET_NAME_BTN_ID);
+        click(NORMAL_SRV_BTN_ID);
+        waitUntilNot(driver -> getText(RESULT_SPAN_ID) == null);
+        Assert.assertEquals("Hello MyName", getText(RESULT_SPAN_ID));
```
mcollovati, line 69: What about making the two service returns a different output, just to be sure that the text of `RESULT_SPAN_ID` has been effectively changed?
For example, instead of just `Hello xyz`, return `Hello from (CDIComponent|NormalScoped), xyz`

vaadin-cdi/src/main/java/com/vaadin/cdi/annotation/CdiComponent.java
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.annotation.*;
+
+/**
+ * This is a {@code @Stereotype} to let the container scan and manage
+ * the instances of the annotated types.
+ * <p>
+ * Annotating the beans with this annotation is necessary when the
+ * {@code bean-discovery-mode="annotated"} is set in the beans.xml
+ * file (implicit bean archive), and in this case CDI can only manage
+ * and inject beans annotated with a "scope" type.
```
mcollovati, line 28: > can only manage and inject beans annotated with a "scope" type.

I cannot understand this statement. Do you mean beans annotated with a "normal scope"? In this case, I would rephrase as `can only manage and inject beans annotated with a bean defining annotations`.

taefi: That statement is coming from https://docs.oracle.com/javaee/7/tutorial/cdi-adv001.htm. But, I agree that rephrasing it to what you proposed can make it clearer.

mcollovati (approved):

PR #438, opened 2023-06-21
fix: fix serialization issues

### LINKED ISSUE #436: RouteScopedContext.NavigationData is not serializable

### LINKED ISSUE #437: Serialization issue with UIEventListener

mcollovati:
## Description

Fixes serialization issues

Fixes #436
Fixes #437

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -185,6 +185,11 @@
                 <version>${mockito.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-inline</artifactId>
+                <version>${mockito.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>com.vaadin</groupId>
                 <artifactId>flow-test-util</artifactId>
                 <version>${vaadin.flow.version}</version>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -184,6 +184,11 @@
                 <artifactId>mockito-core</artifactId>
                 <version>${mockito.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-inline</artifactId>
+                <version>${mockito.version}</version>
+            </dependency>
```
caalador, line 191: Could we just update mockito to 5.4.0 to get rid of inline as that would have it baked in already?

mcollovati: Good point! I'll bump the version

caalador (approved):

PR #439, opened 2023-06-28
fix: fix serialization issues (#438)

### LINKED ISSUE #436: RouteScopedContext.NavigationData is not serializable

### LINKED ISSUE #437: Serialization issue with UIEventListener

mcollovati:
Fixes #436
Fixes #437


czp13 (dismissed):

czp13 (approved):
The covering tests are different because of (I assume) code base differences, RouteScopedContext is new change here, seems safe.  And CdiVaadinServletService.java
are the same.

Also thanks for tackling the extra not needed constructor param here as well.

Good work, it feels safe to merge! 🙇

PR #440, opened 2023-06-28
fix: fix serialization issues (#438)

### LINKED ISSUE #436: RouteScopedContext.NavigationData is not serializable

### LINKED ISSUE #437: Serialization issue with UIEventListener

mcollovati:
Fixes #436
Fixes #437

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
@@ -39,6 +39,7 @@ import com.vaadin.flow.server.SystemMessagesProvider;
 import com.vaadin.flow.server.VaadinService;
 import com.vaadin.flow.server.VaadinServletService;
 import com.vaadin.flow.server.VaadinSession;
+import com.vaadin.flow.server.WrappedSession;
 
 import javax.enterprise.context.spi.Context;
 import javax.enterprise.context.spi.CreationalContext;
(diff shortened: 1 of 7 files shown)


czp13 (approved):

vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
```diff
@@ -146,21 +161,17 @@ public CdiVaadinServlet getServlet() {
      */
     public static class CdiVaadinServiceDelegate implements Serializable {
 
-        private final VaadinService vaadinService;
-
         private transient BeanManager beanManager;
 
         private final UIEventListener uiEventListener;
 
         public CdiVaadinServiceDelegate(VaadinService vaadinService,
```
czp13, line 168: Could we remove from here as well the `vaadinService`, if I am correct it is unused now.

czp13: In that case probably from this line as well: 
`this.delegate = new CdiVaadinServiceDelegate(this, beanManager);`

mcollovati: done

czp13: Thank you vm! 🙇

czp13 (approved):
Checked again, more deeply, code is good, tests are covering changes! Feel safe to merge.

PR #460, opened 2024-10-14
fix: provide default MenuAccessControl instance

mcollovati:
A MenuAccessControl instance is required when operating with MenuConfiguration. This change provides a default implementation if the user project does not define a specialized bean.

### DIFF
--- a/vaadin-cdi/src/test/java/com/vaadin/cdi/CdiInstantiatorTest.java
+++ b/vaadin-cdi/src/test/java/com/vaadin/cdi/CdiInstantiatorTest.java
@@ -180,7 +180,7 @@ public class CdiInstantiatorTest extends AbstractWeldTest {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


vaadin-cdi/src/test/java/com/vaadin/cdi/CdiInstantiatorTest.java
```diff
@@ -162,6 +179,14 @@ public void getI18NProvider_beanEnabled_instanceReturned() {
         Assertions.assertTrue((i18NProvider instanceof I18NTestProvider));
     }
 
+    @Test
+    public void getMenuAccessControl_beanNotProvided_instanceReturned() {
```
tltv, line 183:
```suggestion
    public void getMenuAccessControl_beanEnabled_instanceReturned() {
```

PR #482, opened 2025-07-17
chore: add jcip test dependency

mcollovati:


mcollovati:
jcip-annotation was transitively imported by the license checker. With the latest version, the scope of the dependency has been fixed in LC, so now it needs to be explicitly defined in the project (as it should be).

tltv (approved):

PR #492, opened 2025-11-05
test: upgrade Wildfly to 38

mcollovati:
Also add deployment descriptor to exclude old Jackson version imported by the container.


mcollovati:
> Should this be mentioned in some docs?

It is on the release notes for the moment, but I created this ticket https://github.com/vaadin/docs/issues/4734

PR #498, opened 2026-03-17
feat: Add DetachEvent listener to CdiVaadinServletService

### LINKED ISSUE #497: Add UIDetachEvent CDI event for observing UI being detached
There is both `SessionInitEvent` and `SessionDestroyEvent` available for observing `VaadinSession`, but for observing UI there is only `UIInitEvent`. For sake of symmetry `UIDetachEvent` should be provided.

CDI integration add-on already has UI event handler where this could be added.

TatuLund:
fixes #497

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
@@ -267,7 +267,8 @@ public class CdiVaadinServletService extends VaadinServletService {
     @ListenerPriority(-100) // navigation event listeners are last by default
     private static class UIEventListener
             implements AfterNavigationListener, BeforeEnterListener,
-            BeforeLeaveListener, ComponentEventListener<PollEvent> {
+            BeforeLeaveListener, ComponentEventListener<PollEvent>,
+            ComponentEventListener<DetachEvent> {
 
         private final CdiVaadinServiceDelegate delegate;
 
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
@@ -34,6 +34,7 @@ import com.vaadin.cdi.annotation.VaadinServiceScoped;
 import com.vaadin.cdi.context.VaadinSessionScopedContext;
 import com.vaadin.cdi.util.BeanManagerProvider;
 import com.vaadin.flow.component.ComponentEventListener;
+import com.vaadin.flow.component.DetachEvent;
 import com.vaadin.flow.component.PollEvent;
 import com.vaadin.flow.component.UI;
 import com.vaadin.flow.di.Instantiator;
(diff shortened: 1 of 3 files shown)


vaadin-cdi/src/main/java/com/vaadin/cdi/UIDetachEvent.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package com.vaadin.cdi;
+
+import com.vaadin.flow.component.DetachEvent;
+import com.vaadin.flow.component.UI;
+
+/**
+ * Event fired before a {@link UI} is detached.
+ *
+ * @since
```
mshabarov:
```suggestion
 * @since 16.1
```

vaadin-cdi/src/main/java/com/vaadin/cdi/CdiVaadinServletService.java
```diff
@@ -197,6 +197,8 @@ public void addUIListeners(UI ui) {
             ui.addBeforeLeaveListener(uiEventListener);
             ui.addBeforeEnterListener(uiEventListener);
             ui.addPollListener(uiEventListener);
+            ui.addDetachListener(e -> 
+                    getBeanManager().getEvent().fire(new UIDetachEvent(e)));
```
mshabarov, line 201: Needs a test validating this.

TatuLund: Done

vaadin-cdi/src/test/java/com/vaadin/cdi/CdiVaadinServletServiceTest.java
```diff
@@ -221,6 +232,28 @@ void fireUIInitListeners_serialization_UIserializableAndListenersWork() throws E
...
+        VaadinSession session = new MockVaadinSession(service);
+        session.getLockInstance().lock();
+        try {
+            ui.getInternals().setSession(session);
+            service.fireUIInitListeners(ui);
+        } finally {
+            session.getLockInstance().unlock();
+        }
+
+        ComponentUtil.fireEvent(ui, new DetachEvent(ui));
```
mcollovati: Should be better to use `ui.getInternals().setSession(null)` to simulate UI detach instead of arbitrarily fire the event.

TatuLund: Ok. Done. I added verification of UIInitEvent as well, as the same harness can do that.

mcollovati: Sounds good, but I would then rename the test because it does not focus anymore only on the detach event

TatuLund: Done.

mshabarov (dismissed):

vaadin-cdi/src/test/java/com/vaadin/cdi/CdiVaadinServletServiceTest.java
```diff
@@ -221,6 +245,37 @@ void fireUIInitListeners_serialization_UIserializableAndListenersWork() throws E
         Assertions.assertEquals(ui2, uiListenerEventReceiver.pollEventUI);
     }
 
+    @Test
+    void fireUIInitListeners_uiDetached_UIInitEventAndUIDetachEventCanBeObserved()
```
mcollovati, line 249:
```suggestion
    void fireUIInitListeners_uiAttachedAndDetached_UIeventsCanBeObserved()
```

mcollovati (approved):

PR #500, opened 2026-04-10
chore: add github action workflow

ZheSun88:
- add the github action config 
- performance improved from 30 miutes to 6 minutes via running different servlet in parallel

**when cherry pick, need to updated the branch name, java/maven/node setting and profile for containers**

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -106,8 +106,9 @@ jobs:
         run: tar xf workspace.tar -C ~
 
       - name: Set TB License
+        env:
+          TB_LICENSE: ${{secrets.TB_LICENSE}}
         run: |
-          TB_LICENSE=${{secrets.TB_LICENSE}}
           mkdir -p ~/.vaadin/
           echo '{"username":"'`echo $TB_LICENSE | cut -d / -f1`'","proKey":"'`echo $TB_LICENSE | cut -d / -f2`'"}' > ~/.vaadin/proKey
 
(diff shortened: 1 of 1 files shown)


.github/workflows/validation.yml
```diff
@@ -0,0 +1,165 @@
...
+        uses: actions/download-artifact@v5
+        with:
+          name: saved-workspace
+
+      - name: Restore workspace
+        run: tar xf workspace.tar -C ~
+
+      - name: Set TB License
+        run: |
+          TB_LICENSE=${{secrets.TB_LICENSE}}
```
mcollovati: Shouldn't this be an `env` entry instead?

.github/workflows/validation.yml
```diff
@@ -0,0 +1,165 @@
...
+        run: |
+          find . -name surefire-reports -type d \
+            | tar -czf tests-report-${{ matrix.profile }}.tgz -T -
+
+      - name: Upload test reports
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v5
+        with:
+          name: tests-output-${{ matrix.profile }}
+          path: tests-report-${{ matrix.profile }}.tgz
```
mcollovati: For troubleshooting, it would be good to also upload error screenshots.

.github/workflows/validation.yml
ZheSun88: updated

.github/workflows/validation.yml
ZheSun88: updated

.github/workflows/validation.yml
```diff
@@ -0,0 +1,169 @@
...
+      actions: write
+      checks: write
+      pull-requests: write
+    steps:
+      - name: Merge test artifacts
+        uses: actions/upload-artifact/merge@v5
+        with:
+          name: tests-output
+          pattern: tests-output-*
+          delete-merged: true
```
mcollovati, line 152: How will this merge action work with error screenshots?
I mean, if the exactly same test fails on different containers for different reasons, will there be only one screenshot in the attached artifact?

ZheSun88: good catch.. let us add the `separate-directories: true`

mcollovati (approved):

PR #507, opened 2026-05-21
Fix: #506 - VaadinSessionScopedContext was no longer active

### LINKED ISSUE #506: Commit 0d3aeeb causes many neccessary changes to use the VaadinSessionScoped Context again
The Issue is that the Following line was adapted to check for the Session-Lock. 

Image

This should not happen here. Instead the Places where the Session is actually need to determine the Lock-State.
Otherwise you'll always receive a ContextNotActive exception. In our Project we've like 4500 Places to fix - This is nothing for a minor patch.
Since CDI should be automatically managed, the lock needs to be taken, if necessary.

OriginalFelix:
## Description

In a Multi-Threaded Application 1 View with 10 Threads for Example, with threads containing the UI and VaadinSession, the Threads could access the CDI-Scopes before the changes breaking changes in the isActive method.
These changes were reverted and instead the findContextualStorage Method was adapted to dynamically take the lock, if not already running on a UI-Thread. This simplifies the adaptation for the customers, due to them not having to change every single line of code where they accessing the VaadinSessionScoped beans.

Fixes #506

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [X] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
@@ -43,7 +43,6 @@ import com.vaadin.cdi.context.VaadinSessionScopedContext;
 import com.vaadin.cdi.util.AbstractContext;
 import com.vaadin.cdi.util.BeanProvider;
 import com.vaadin.cdi.util.DependentProvider;
-import com.vaadin.flow.server.ServiceInitEvent;
 
 /**
  * CDI Extension needed to register Vaadin scopes to the runtime.
@@ -67,11 +66,7 @@ public class VaadinExtension implements Extension {
         uiScopedContext = new UIScopedContext(beanManager);
         routeScopedContext = new RouteScopedContext(beanManager);
         addContext(afterBeanDiscovery, serviceScopedContext, null);
-        VaadinSessionScopedContext sessionScopedContext = new VaadinSessionScopedContext(beanManager);
-        afterBeanDiscovery.<ServiceInitEvent>addObserverMethod()
-            .observedType(ServiceInitEvent.class)
-            .notifyWith(instance -> sessionScopedContext.initializeActivationPolicy(instance.getEvent()));
-        addContext(afterBeanDiscovery,sessionScopedContext, null);
+        addContext(afterBeanDiscovery,new VaadinSessionScopedContext(beanManager), null);
         addContext(afterBeanDiscovery, uiScopedContext, NormalUIScoped.class);
         addContext(afterBeanDiscovery, routeScopedContext,
                 NormalRouteScoped.class);
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
@@ -35,8 +35,6 @@ import com.vaadin.cdi.DeploymentValidator.BeanInfo;
 import com.vaadin.cdi.annotation.NormalRouteScoped;
 import com.vaadin.cdi.annotation.NormalUIScoped;
 import com.vaadin.cdi.annotation.VaadinServiceEnabled;
-import com.vaadin.cdi.annotation.VaadinSessionScopeActivationPolicy;
-import com.vaadin.cdi.annotation.VaadinSessionScopeActivationPolicy.Policy;
 import com.vaadin.cdi.context.ContextWrapper;
 import com.vaadin.cdi.context.RouteScopedContext;
 import com.vaadin.cdi.context.UIScopedContext;
@@ -45,11 +43,7 @@ import com.vaadin.cdi.context.VaadinSessionScopedContext;
 import com.vaadin.cdi.util.AbstractContext;
 import com.vaadin.cdi.util.BeanProvider;
 import com.vaadin.cdi.util.DependentProvider;
-import com.vaadin.flow.component.page.AppShellConfigurator;
-import com.vaadin.flow.server.AppShellRegistry;
 import com.vaadin.flow.server.ServiceInitEvent;
-import com.vaadin.flow.server.VaadinContext;
-import com.vaadin.flow.server.VaadinService;
 
 /**
  * CDI Extension needed to register Vaadin scopes to the runtime.
@@ -61,10 +55,6 @@ public class VaadinExtension implements Extension {
     private RouteScopedContext routeScopedContext;
     private Set<BeanInfo> beanInfoSet = new HashSet<>();
 
-    /**
-     * The current VaadinSessionScopeActivationPolicy.
-     */
-    private static Policy vaadinSessionScopeActivationPolicy = VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
 
     private void storeBeanValidationInfo(@Observes ProcessBean processBean) {
         beanInfoSet.add(new BeanInfo(processBean.getBean(),
@@ -74,53 +64,19 @@ public class VaadinExtension implements Extension {
     private void addContexts(@Observes AfterBeanDiscovery afterBeanDiscovery,
             BeanManager beanManager) {
         serviceScopedContext = new VaadinServiceScopedContext(beanManager);
-        afterBeanDiscovery.<ServiceInitEvent>addObserverMethod()
-            .observedType(ServiceInitEvent.class)
-            .notifyWith(instance -> this.onServiceInit(instance.getEvent()));
         uiScopedContext = new UIScopedContext(beanManager);
         routeScopedContext = new RouteScopedContext(beanManager);
         addContext(afterBeanDiscovery, serviceScopedContext, null);
-        addContext(afterBeanDiscovery,
-                new VaadinSessionScopedContext(beanManager), null);
+        VaadinSessionScopedContext sessionScopedContext = new VaadinSessionScopedContext(beanManager);
+        afterBeanDiscovery.<ServiceInitEvent>addObserverMethod()
+            .observedType(ServiceInitEvent.class)
+            .notifyWith(instance -> sessionScopedContext.initializeActivationPolicy(instance.getEvent()));
+        addContext(afterBeanDiscovery,sessionScopedContext, null);
         addContext(afterBeanDiscovery, uiScopedContext, NormalUIScoped.class);
         addContext(afterBeanDiscovery, routeScopedContext,
                 NormalRouteScoped.class);
     }
 
-    /**
-     * Called when the VaadinService is initialized.
-     * @param event the ServiceInitEvent
-     */
-    private void onServiceInit(final ServiceInitEvent event) {
-        vaadinSessionScopeActivationPolicy = this.determineVaadinSessionScopeActivationPolicy(event.getSource());
-    }
-
-    /**
-     * Determine the VaadinSessionScopeActivationPolicy for the current VaadinService.
-     * @param vaadinService the current VaadinService
-     * @return the determined policy
-     */
-    private Policy determineVaadinSessionScopeActivationPolicy(final VaadinService vaadinService) {
-        if (vaadinService == null) {
-            return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
-        }
-        final VaadinContext context = vaadinService.getContext();
-        if (context == null) {
-            return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
-        }
-        final AppShellRegistry registry = AppShellRegistry.getInstance(context);
-        if (registry == null) {
-            return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
-        }
-        final Class<? extends AppShellConfigurator> configurator = registry.getShell();
-        if (configurator == null) {
-            return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
-        }
-        if (configurator.isAnnotationPresent(VaadinSessionScopeActivationPolicy.class)) {
-            return configurator.getAnnotation(VaadinSessionScopeActivationPolicy.class).value();
-        }
-        return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
-    }
 
     // Validate annotated executor
 
@@ -170,14 +126,6 @@ public class VaadinExtension implements Extension {
 
     private static Logger getLogger() {
         return LoggerFactory.getLogger(VaadinExtension.class);
-    }
-
-    /**
-     * Get the current VaadinSessionScopeActivationPolicy.
-     * @return the current VaadinSessionScopeActivationPolicy
-     */
-    public static Policy getVaadinSessionScopeActivationPolicy() {
-        return vaadinSessionScopeActivationPolicy;
     }
 
 }
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
@@ -54,7 +54,6 @@ public class VaadinExtension implements Extension {
     private RouteScopedContext routeScopedContext;
     private Set<BeanInfo> beanInfoSet = new HashSet<>();
 
-
     private void storeBeanValidationInfo(@Observes ProcessBean processBean) {
         beanInfoSet.add(new BeanInfo(processBean.getBean(),
                 processBean.getAnnotated()));
@@ -66,12 +65,12 @@ public class VaadinExtension implements Extension {
         uiScopedContext = new UIScopedContext(beanManager);
         routeScopedContext = new RouteScopedContext(beanManager);
         addContext(afterBeanDiscovery, serviceScopedContext, null);
-        addContext(afterBeanDiscovery,new VaadinSessionScopedContext(beanManager), null);
+        addContext(afterBeanDiscovery,
+                new VaadinSessionScopedContext(beanManager), null);
         addContext(afterBeanDiscovery, uiScopedContext, NormalUIScoped.class);
         addContext(afterBeanDiscovery, routeScopedContext,
                 NormalRouteScoped.class);
     }
-
 
     // Validate annotated executor
 
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
@@ -35,6 +35,8 @@ import com.vaadin.cdi.DeploymentValidator.BeanInfo;
 import com.vaadin.cdi.annotation.NormalRouteScoped;
 import com.vaadin.cdi.annotation.NormalUIScoped;
 import com.vaadin.cdi.annotation.VaadinServiceEnabled;
+import com.vaadin.cdi.annotation.VaadinSessionScopeActivationPolicy;
+import com.vaadin.cdi.annotation.VaadinSessionScopeActivationPolicy.Policy;
 import com.vaadin.cdi.context.ContextWrapper;
 import com.vaadin.cdi.context.RouteScopedContext;
 import com.vaadin.cdi.context.UIScopedContext;
@@ -43,6 +45,11 @@ import com.vaadin.cdi.context.VaadinSessionScopedContext;
 import com.vaadin.cdi.util.AbstractContext;
 import com.vaadin.cdi.util.BeanProvider;
 import com.vaadin.cdi.util.DependentProvider;
+import com.vaadin.flow.component.page.AppShellConfigurator;
+import com.vaadin.flow.server.AppShellRegistry;
+import com.vaadin.flow.server.ServiceInitEvent;
+import com.vaadin.flow.server.VaadinContext;
+import com.vaadin.flow.server.VaadinService;
 
 /**
  * CDI Extension needed to register Vaadin scopes to the runtime.
@@ -54,6 +61,11 @@ public class VaadinExtension implements Extension {
     private RouteScopedContext routeScopedContext;
     private Set<BeanInfo> beanInfoSet = new HashSet<>();
 
+    /**
+     * The current VaadinSessionScopeActivationPolicy.
+     */
+    private static Policy vaadinSessionScopeActivationPolicy = VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
+
     private void storeBeanValidationInfo(@Observes ProcessBean processBean) {
         beanInfoSet.add(new BeanInfo(processBean.getBean(),
                 processBean.getAnnotated()));
@@ -62,6 +74,9 @@ public class VaadinExtension implements Extension {
     private void addContexts(@Observes AfterBeanDiscovery afterBeanDiscovery,
             BeanManager beanManager) {
         serviceScopedContext = new VaadinServiceScopedContext(beanManager);
+        afterBeanDiscovery.<ServiceInitEvent>addObserverMethod()
+            .observedType(ServiceInitEvent.class)
+            .notifyWith(instance -> this.onServiceInit(instance.getEvent()));
         uiScopedContext = new UIScopedContext(beanManager);
         routeScopedContext = new RouteScopedContext(beanManager);
         addContext(afterBeanDiscovery, serviceScopedContext, null);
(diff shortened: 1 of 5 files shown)


vaadin-cdi/src/main/java/com/vaadin/cdi/annotation/VaadinSessionScopeActivationPolicy.java
```diff
@@ -0,0 +1,22 @@
+package com.vaadin.cdi.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface VaadinSessionScopeActivationPolicy {
```
mcollovati, line 40: Needs some class level Javadoc, describing the purpose of this annotation and how it is supposed to be used. Adding a short code snippet could also be useful.

vaadin-cdi/src/test/java/com/vaadin/cdi/context/SessionContextTest.java
mcollovati, line 1: We need to improve tests by adding cases for both strict and non-strict mode.

vaadin-cdi/src/main/java/com/vaadin/cdi/annotation/VaadinSessionScopeActivationPolicy.java
```diff
@@ -0,0 +1,22 @@
...
+	/**
+	 * Default value for {@link #strict()}
+	 */
+	boolean DEFAULT_STRICT = false;
+
+	/**
+	 * Determines if the session scope should check the session lock to activate the Scope.
+	 * @return true if strict session locking should be enabled, false otherwise
+	 */
+	boolean strict() default DEFAULT_STRICT;
```
mcollovati: Random though: what about using an enum instead of a boolean (`Policy.STRICT`, `Policly.RELAXED`)?
This could probably help also in `shouldUseStrictChecks`, by having a single `AtomicReference` instead of two fields to handle.

In addition, we could move the `isActive()` logic here, defining a `isActive(VaadinSession)` method.

Not a blocker anyway. Feel free to ignore this comment.

mcollovati (changes requested):
I added a couple of comments. Overall looks good; I'm just a bit sad that we couldn't move the detection to the `VaadinExtension` class to execute it eagerly.

mcollovati:
Maybe a stupid idea, but what about programmatically defining an observer for `ServiceInitEvent` in `VaadinExtension`?

```java
        var serviceScopedContext = new VaadinServiceScopedContext(beanManager);
        afterBeanDiscovery.<ServiceInitEvent>addObserverMethod()
                    .observedType(ServiceInitEvent.class)
                    .notifyWith(instance -> {
                        sessionContext.onServiceInit(instance.getEvent());
                    });
```

This could be added to `VaadinExtension.addContexts()`; the event is fired before any VaadinSession is created, so it should be safe. `VaadinSessionScopedContext.onServiceInit(...)` could have a simplified version of `shouldUseStrictChecks`, since at this point we are sure `VaadinService` and `VaadinContext` are not null

vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
```diff
@@ -72,6 +87,41 @@ private void addContexts(@Observes AfterBeanDiscovery afterBeanDiscovery,
...
+    private void onServiceInit(final ServiceInitEvent event) {
+        vaadinSessionScopeActivationPolicy = this.determineVaadinSessionScopeActivationPolicy(event.getSource());
+    }
+
+    /**
+     * Determine the VaadinSessionScopeActivationPolicy for the current VaadinService.
+     * @param vaadinService the current VaadinService
+     * @return the determined policy
+     */
+    private Policy determineVaadinSessionScopeActivationPolicy(final VaadinService vaadinService) {
```
mcollovati: I would prefer to have this logic in `VaadinSessionScopedContext`. The synthetic observer could directly call a method on `VaadinSessionScopedContext` instance created in `addContexts`, instead of using `VaadinExtension` as a value holder.

OriginalFelix: I've adapted the PR :)

vaadin-cdi/src/main/java/com/vaadin/cdi/context/VaadinSessionScopedContext.java
```diff
@@ -49,28 +64,104 @@ protected ContextualStorage getContextualStorage(Contextual<?> contextual, boole
...
+     * Determines if the context is active based on the current session and activation policy.
+     * @return true if the context is active, false otherwise
+     */
     @Override
     public boolean isActive() {
-        VaadinSession session = VaadinSession.getCurrent();
-        return session != null && session.hasLock();
+        final VaadinSession session = VaadinSession.getCurrent();
+        if (VaadinSessionScopedContext.getActivationPolicy() == null && session != null) {
+            activationPolicy.compareAndSet(null, determineVaadinSessionScopeActivationPolicy(session.getService()));
```
mcollovati: Why use a static field to store the policy?
Also, why we should call `determineVaadinSessionScopeActivationPolicy` again? The  `VaadinService` init event should always be fired before any VaadinSession gets created.
Can you elaborate on this?

OriginalFelix: @mcollovati yes, the problem here is that the Event is only fired, when the CDI-Plugin is deployed to a Servlet Container, but not inside the Unit-Tests. Maybe i'm missing something there?

mcollovati: OK, so it's mostly an issue in how the tests are constructed, right?
If so, let me check.
The other question is if we should assume a single `VaadinService` or multiple ones (even if I guess this is an extremely rare if not inexistent scenario)

OriginalFelix: Yes.
Is it possible to have multiple VaadinServices in the same war?

mcollovati: Potentially, yes, by having multiple `VaadinServlet`s. Actually I don't remember if this configuration is supported or not.
Anyway, I'll propose you another approach that should handle also that case (sorry for being so picky :stuck_out_tongue_winking_eye:)

The idea is to follow the same pattern applied to get AppShellRegistry (the following code is not tested, I just put down the ide):

1. Define a `VaadinSessionActivationPolicyHolder` class with a `VaadinSessionScopeActivationPolicy.Policy get(VaadinService vaadinService)` method that resembles `AppShellRegistry.getInstance()`: get `VaadinContext` from service and store the policy in a private wrapper class

```java
class VaadinSessionActivationPolicyHolder {

    public static VaadinSessionScopeActivationPolicy.Policy get(
            VaadinService vaadinService) {
        VaadinContext context = vaadinService.getContext();
        synchronized (context) { // NOSONAR
            PolicyWrapper attribute = context.getAttribute(PolicyWrapper.class);
            if (attribute == null) {
                VaadinSessionScopeActivationPolicy.Policy policy = determinePolicy(
                        vaadinService);
                attribute = new PolicyWrapper(policy);
                context.setAttribute(attribute);
            }
            return attribute.policy;
        }
    }

    private static VaadinSessionScopeActivationPolicy.Policy determinePolicy(
            VaadinService vaadinService) {
        // current logic
    }

    private record PolicyWrapper(VaadinSessionScopeActivationPolicy.Policy policy) {
    }
}
```

2. In `VaadinServiceScopedContext` add a `ServiceInitEvent` observer and call `VaadinSessionActivationPolicyHolder.get(...)`. This way we will have the policy stored for every `VaadinService` that is created.

3. In `VaadinSessionScopedContext.isActive` simply call `VaadinSessionActivationPolicyHolder.get(session.getService())` and use that value


How does this sound?

EDIT: we need to carefully verify the synchronized stuff

OriginalFelix: The VaadinContext isn't mocked - the tests pass, but they always recreate the PolicyHolder

OriginalFelix: And the synchronized Block will cause issues with concurrency even if we use the internal method context.getAttribute(PolicyWrapper.class, () -> new PolicyWrapper(determinePolicy(vaadinService)));

mcollovati: I guess we can add a package-private `reset()` method on `VaadinSessionActivationPolicyHolder`.
Then the tests can call `AppShellRegistry.getInstance(vaadinContext).setShell(StrictAppShell.class)` (and appShellRegistry.reset()` after the test, in a finally block)

mcollovati: I'm not familiar with these tests. Let me double-check

OriginalFelix: The other Problem is once we call getAttribute() on the VaadinContext when a Session is created, there's a chance the Session could be deadlocked, when 2 Sessions are created at the same time. Due to the internal functionality of the method in the VaadinContext which uses a synchronized Block :/
And this could even happen when CDI events are fired - We could reduce the chance of this happening by caching the Value in the VaadinSession, but this would complicate things even more...
In my opinion the static variable is the best Option, to prevent deadlocks and keep performance up :)

mcollovati: We need a small change in `TestCdiVaadinServletService`, something like this

```java
        Map<String, Object> contextAttributes = new HashMap<>();
        when(servletcontext.getAttribute(anyString())).then(i -> contextAttributes.get(i.<String>getArgument(0)));
        doAnswer(i -> {
            contextAttributes.put(i.getArgument(0), i.getArgument(1));
            return null;
        }).when(servletcontext).setAttribute(anyString(), any());
        Lookup lookup = mock(Lookup.class);
        contextAttributes.put(Lookup.class.getName(), lookup);
        //when(servletcontext.getAttribute(Lookup.class.getName()))
        //        .thenReturn(lookup);
        when(lookup.lookup(ApplicationConfigurationFactory.class))
                .thenReturn(context -> mock(ApplicationConfiguration.class));
```

I can take care of fixing the tests after you push your changes

mcollovati: > The other Problem is once we call getAttribute() on the VaadinContext when a Session is created

When is `VaadinContext.getAttribute()` called in VaadinSession creation? I'm missing something here

mcollovati: Or better, how is VaadinSession creation related to the policy initialization?

OriginalFelix: The problem is that the isActive is calling the VaadinSessionActivationPolicyHolder.get(), which needs to access the VaadinContext, which internally uses synchronized to lock.

mcollovati: I see, but I still don't understand how session creation is involved

OriginalFelix: The SessionInitEvent is fired -> Every CDI-Event that is inside the Session Scope is concerned as well

mcollovati: To be honest, I could not see the dead look either. If the method is called from two threads, they contend the same object monitor. I'm still missing something?

Would the following be safer?

```java
  class VaadinSessionActivationPolicyHolder {

      static VaadinSessionScopeActivationPolicy.Policy get(
              VaadinService vaadinService) {
          VaadinContext context = vaadinService.getContext();

          PolicyWrapper wrapper = context.getAttribute(PolicyWrapper.class);
          if (wrapper != null) {
              return wrapper.policy;
          }

          VaadinSessionScopeActivationPolicy.Policy policy =
                  determinePolicy(vaadinService);

          return context.getAttribute(PolicyWrapper.class,
                  () -> new PolicyWrapper(policy)).policy;
      }

  }
```

OriginalFelix: as long as the Object reference (VaadinContext) is the same - the internal synchronization of the context.getAttribute(PolicyWrapper.class); function will cause monitoring threads and possible deadlocks - the VaadinContext itself contains the problem :)

OriginalFelix: Image

mcollovati: What's your proposal? A single static variable would not work for multiple VaadinService. And I guess we will not hold references to VaadinService.

OriginalFelix: @mcollovati I've added some lines - We might not even need the initialization inside the Service, what about the first Session caches the Value in the VaadinContext via the given functions and copies it down into the Session to prevent synchronized locks :)

mcollovati: But then the session will be locked every time `isActive()` is called

OriginalFelix: That's correct. Either this or we go back to the static variable - the synchronized lock risk is just too high.

mcollovati: And if we just add a ConcurrenHashMap<String,Policy> to `VaadinSessionActivationPolicyHolder` and use `vaadinService.getServiceName()` as the key? It basically is the same key used by `VaadinServiceScopedContext`

mcollovati: Or we can mimic/reuse what's done in `VaadinServiceScopedContext.getContextualStorage` to get the key

OriginalFelix: that would work

OriginalFelix: i'll adapt it

OriginalFelix: Done

mcollovati: Looks good, thank you!
I guess we can remove some redundant null checks from `VaadinSessionScopedContext` if `VaadinSessionActivationPolicyHolder` already handles them. And most likely we can drop `getActivationPolicy` as well.
Also, `determinePolicy` is never called with a null parameter, so that's another check that can be removed.


I wonder if `VaadinSessionActivationPolicyHolder` could be handled observice `ServiceInitEvent` and `ServiceDestroyEvent` events. But that's not a blocker.

vaadin-cdi/src/main/java/com/vaadin/cdi/util/VaadinSessionActivationPolicyHolder.java
```diff
@@ -0,0 +1,89 @@
...
+import com.vaadin.cdi.annotation.VaadinSessionScopeActivationPolicy;
+import com.vaadin.cdi.annotation.VaadinSessionScopeActivationPolicy.Policy;
+import com.vaadin.flow.component.page.AppShellConfigurator;
+import com.vaadin.flow.server.AppShellRegistry;
+import com.vaadin.flow.server.VaadinContext;
+import com.vaadin.flow.server.VaadinService;
+import com.vaadin.flow.server.VaadinSession;
+
+/**
+ * Utility class to get the VaadinSessionScopeActivationPolicy for the current VaadinService.
```
mcollovati:
```suggestion
 * Utility class to get the VaadinSessionScopeActivationPolicy for a VaadinService.
```

vaadin-cdi/src/main/java/com/vaadin/cdi/util/VaadinSessionActivationPolicyHolder.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public class VaadinSessionActivationPolicyHolder {
+
+	/**
+	 * Caches the Policies for each VaadinService.
+	 */
+	private static final ConcurrentHashMap<String, PolicyWrapper> policyCache = new ConcurrentHashMap<>();
```
mcollovati: `PolicyWrapper` is useless with this design, we can directly store the `Policy`

vaadin-cdi/src/main/java/com/vaadin/cdi/annotation/VaadinSessionScopeActivationPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface VaadinSessionScopeActivationPolicy {
+
+	/**
+	 * Activation policies for the VaadinSessionScopedContext.
+	 */
+	enum Policy {
+		STRICT,
+		LENIENT
```
mcollovati, line 53: Entries needs Javadoc as well

vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
```diff
@@ -54,6 +54,7 @@ public class VaadinExtension implements Extension {
     private RouteScopedContext routeScopedContext;
     private Set<BeanInfo> beanInfoSet = new HashSet<>();
 
+
```
mcollovati:
```suggestion
```

vaadin-cdi/src/main/java/com/vaadin/cdi/VaadinExtension.java
```diff
@@ -65,13 +66,13 @@ private void addContexts(@Observes AfterBeanDiscovery afterBeanDiscovery,
...
         addContext(afterBeanDiscovery, serviceScopedContext, null);
-        addContext(afterBeanDiscovery,
-                new VaadinSessionScopedContext(beanManager), null);
+        addContext(afterBeanDiscovery,new VaadinSessionScopedContext(beanManager), null);
         addContext(afterBeanDiscovery, uiScopedContext, NormalUIScoped.class);
         addContext(afterBeanDiscovery, routeScopedContext,
                 NormalRouteScoped.class);
     }
 
+
```
mcollovati:
```suggestion
```

mcollovati (changes requested):

mcollovati:
Got the following error on shutdown

```
17:05:20,697 ERROR [com.vaadin.flow.server.DefaultErrorHandler] (ServerService Thread Pool -- 82) Unexpected error: ServletConfig has not been initialized: java.lang.IllegalStateException: ServletConfig has not been initialized
        at jakarta.servlet.api@6.0.0//jakarta.servlet.GenericServlet.getServletName(GenericServlet.java:244)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinServletService.getServiceName(VaadinServletService.java:199)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.util.VaadinSessionActivationPolicyHolder.get(VaadinSessionActivationPolicyHolder.java:34)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.util.VaadinSessionActivationPolicyHolder.get(VaadinSessionActivationPolicyHolder.java:47)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.context.VaadinSessionScopedContext.getActivationPolicy(VaadinSessionScopedContext.java:151)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.context.VaadinSessionScopedContext.isActive(VaadinSessionScopedContext.java:110)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.context.ContextWrapper.isActive(ContextWrapper.java:60)
        at org.jboss.weld.core@5.1.6.Final//org.jboss.weld.manager.BeanManagerImpl.internalGetContext(BeanManagerImpl.java:681)
        at org.jboss.weld.core@5.1.6.Final//org.jboss.weld.manager.BeanManagerImpl.getContext(BeanManagerImpl.java:652)
        at org.jboss.weld.core@5.1.6.Final//org.jboss.weld.util.ForwardingBeanManager.getContext(ForwardingBeanManager.java:178)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.util.ContextUtils.isContextActive(ContextUtils.java:59)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.util.ContextUtils.isContextActive(ContextUtils.java:45)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.context.VaadinSessionScopedContext.guessContextIsUndeployed(VaadinSessionScopedContext.java:138)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.cdi.CdiVaadinServletService$CdiVaadinServiceDelegate.sessionDestroy(CdiVaadinServletService.java:232)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinService.lambda$fireSessionDestroy$6(VaadinService.java:896)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
        at java.base/java.util.stream.Streams$ConcatSpliterator.forEachRemaining(Streams.java:735)
        at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinService.lambda$fireSessionDestroy$9c853e43$1(VaadinService.java:894)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinService.runPendingAccessTasks(VaadinService.java:2340)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinSession.unlock(VaadinSession.java:790)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinService.ensureAccessQueuePurged(VaadinService.java:2304)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinService.accessSession(VaadinService.java:2271)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinSession.access(VaadinSession.java:1095)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinService.fireSessionDestroy(VaadinService.java:865)
        at deployment.cdi-vs-scope-reproducer-1.0-SNAPSHOT.war//com.vaadin.flow.server.VaadinSession.valueUnbound(VaadinSession.java:220)
        at io.undertow.servlet@2.3.23.Final//io.undertow.servlet.core.SessionListenerBridge.attributeRemoved(SessionListenerBridge.java:145)
```

mcollovati:
I guess this only needs the service name to be computed outside the destroy listener.

EDIT: nope. That is only one point; `get` method needs to be fixed as well

mcollovati:
I did a quick test making `VaadinSessionActivationPolicyHolder` an application scope bean, handling the cache in service init and destroy, and having `get` only perform lookup (with the addition of a `return if cache is empty` guard).
This seems to work fine. But feel free to handle it differently.

EDIT: `return if cache is empty`: that's actually a bit stupid since it could miserably fail if there are more than one VaadinService

OriginalFelix:
The last one was actually tricky! the ServletName requires the Servlet to be Initialized... The solution was to check for the State of the VaadinService and Cache in name into a final String - So it's being stored once it's initialized....

mcollovati:
> The last one was actually tricky! the ServletName requires the Servlet to be Initialized... The solution was to check for the State of the VaadinService and Cache in name into a final String - So it's being stored once it's initialized....

Well done!
The alternative could be to cache the service name in `CdiVaadinServletService`, but of course it would fail with other implementations.

I'll make a final review on Monday, but I think we are close to having this ready.
Thanks for your patience! :medal_sports:

vaadin-cdi/src/main/java/com/vaadin/cdi/util/VaadinSessionActivationPolicyHolder.java
```diff
@@ -0,0 +1,107 @@
...
+	/**
+	 * Get the VaadinSessionScopeActivationPolicy for the current VaadinService.
+	 * @param vaadinService the current VaadinService
+	 * @return the determined policy
+	 */
+	public static Policy get(final VaadinService vaadinService) {
+		if (vaadinService == null) {
+			return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
+		}
+		final String servletName = isServiceInitialized(vaadinService) ? vaadinService.getServiceName() : null;
```
mcollovati: nit: serviceName would be a better name

vaadin-cdi/src/main/java/com/vaadin/cdi/util/VaadinSessionActivationPolicyHolder.java
```diff
@@ -0,0 +1,107 @@
...
+
+	/**
+	 * Determine the VaadinSessionScopeActivationPolicy for the current VaadinService.
+	 * @param vaadinService the current VaadinService
+	 * @return the determined policy
+	 */
+	private static Policy determinePolicy(final VaadinService vaadinService) {
+		if (vaadinService == null) {
+			return VaadinSessionScopeActivationPolicy.DEFAULT_POLICY;
+		}
```
mcollovati: Can be removed. This private method always get a VaadinService instance
```suggestion
```

mcollovati (changes requested):

mcollovati:
Closing as superseded by #515 with a different approach.

PR #512, opened 2026-05-25
Add gha validation for 4.0

ZheSun88:


ZheSun88:
● Two failures in that run:

  1. Payara — POM error: line 520, jakarta.platform:jakarta.jakartaee-api declared without a version. Maven can't even build the POM. This is a real bug in the payara profile.
  2. WildFly — 22 test errors: real CDI Weld WELD-001408: Unsatisfied dependencies for type CDIViewProvider at @Inject private com.vaadin.cdi.CDINavigator.cdiViewProvider. Tests run but Weld can't satisfy injections. This isn't
   a CI/workflow issue — the new pipeline is surfacing pre-existing test-deployment problems.

PR #515, opened 2026-05-26
refactor: extract `@VaadinSessionScoped` activation into managed bean

mcollovati:
Move the activation predicate and storage-access logic out of `VaadinSessionScopedContext` into a nested `ContextualStorageManager` managed bean. The context becomes a thin delegator. As a side effect, applications may replace the bean via CDI `@Specializes` to adjust behavior — though the manager is intentionally undocumented and overriding it is at the integrator's own risk.

Adds integration tests covering the strict default and the lenient `@Specializes` scenario from a background thread that does not hold the session lock.

References #506

### DIFF
--- a/vaadin-cdi-itest/src/main/java/com/vaadin/cdi/itest/sessioncontextspecializes/SessionContextSpecializesView.java
+++ b/vaadin-cdi-itest/src/main/java/com/vaadin/cdi/itest/sessioncontextspecializes/SessionContextSpecializesView.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2000-2018 Vaadin Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.vaadin.cdi.itest.sessioncontextspecializes;
+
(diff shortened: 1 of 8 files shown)


mshabarov:
Let's also follow the internal discussion in Slack.

vaadin-cdi-itest/src/main/java/com/vaadin/cdi/itest/sessioncontextspecializes/SessionContextSpecializesView.java
```diff
@@ -0,0 +1,126 @@
...
+                    } catch (RuntimeException e) {
+                        counter.increment(UNEXPECTED_ERROR_COUNT);
+                    }
+                } finally {
+                    VaadinSession.setCurrent(previous);
+                }
+            }, "background-event-fire");
+            thread.start();
+            try {
+                thread.join(5000);
```
heruan: Wouldn't this hold the lock for the whole 5 secs?

heruan (changes requested):

vaadin-cdi-itest/src/main/java/com/vaadin/cdi/itest/sessioncontextspecializes/SessionContextSpecializesView.java
mcollovati: This is a dirty hack to avoid deadlocks in the test. Let me rewrite it in a better way

mcollovati: Refactored view and test

heruan (approved):

nico-fri:
Is there a plan for when the new Vaadin CDI 16.x version will be released, and will it be included in that release, since only 15.2 is listed?

ZheSun88:
Hi @nico-fri , 

Vaadin CDI 16.0.2 has been released with this fix, you can get it from maven central already.
CDI 15.2.3 should be in maven central quite soon. 

Vaadin Platform with the cooresponding CDI addon version will be released in the next weeks.

🙇

PR #527, opened 2026-07-29
fix: bind route scope to window name only for `@PreserveOnRefresh`

### LINKED ISSUE #526: @RouteScoped route target instance is shared across UIs with the same window.name and reattached after destruction ("Can't move a node from one state tree to another")
## Description

A route target annotated with `@RouteScoped` (pseudo-scope, no client proxy) is handed out to a **second UI** when that UI reports the same `window.name` as the UI that created the bean — even though no `@PreserveOnRefresh` annotation is present anywhere.

Because the component instance still belongs to the first UI's state tree, showing it as the route target of the second UI fails with:

```
java.lang.IllegalStateException:
Can't move a node from one state tree to another.
If this is intentional, first remove the node from its current state tree by calling removeFromTree.
Offending component: created: Component 'TestView' at 'TestView.java' (<init> LINE 19),
attached: Component 'TestView' at 'TestView.java' (<init> LINE 19)
```

In the refresh variant the bean's route scope is even destroyed first — `@PreDestroy` runs — and the destroyed instance is subsequently used as the route target of the new UI (no new constructor / `@PostConstruct` invocation).

Two UIs share a `window.name` in everyday situations:

- the user duplicates a browser tab (Chrome copies `window.name` into the duplicate),
- the browser restores the previous session ("Continue where you left off" keeps `window.name`),
- the user refreshes several times in quick succession, so a new UI is initialized while older UIs of the same window are still awaiting cleanup.

The same code worked correctly with Vaadin 24 and Vaadin CDI 15.2.1.

The behavior surfaces in Vaadin 25 through the combination of two changes:

- https://github.com/vaadin/flow/pull/22719 — browser details (including the `v-wn` window name) are now collected synchronously with the `?v-r=init` request and `UIInternals.getExtendedClientDetails()` never returns null, so a second UI of the same browser window knows its window name immediately;
- https://github.com/vaadin/cdi/pull/493 — the route-scope logic was adapted to that Flow change by detecting fetched client details via `ExtendedClientDetails.windowName`.

The window-name-based storage keying and preservation that these changes activate exists to support `@PreserveOnRefresh` — but it is applied **unconditionally**. Neither our applications nor the reproducer use `@PreserveOnRefresh` anywhere, yet the route-scope storage is still keyed and preserved per window name.

## Environment

- Vaadin platform: 25.2.3 (Flow `flow-server` 25.2.4)
- Vaadin CDI: 16.1.0 (latest; the relevant code is identical in 16.0.0, 16.0.1, 16.0.2)
- CDI implementation: Weld `weld-servlet-core` 5.1.2.Final packaged in the WAR (also reproduced with 5.1.7.Final)
- Servlet container: Apache Tomcat 11.0.24 (Servlet 6.1)
- Servlet: `com.vaadin.cdi.CdiVaadinServlet` mapped to `/*` in `web.xml`
- `beans.xml`: `bean-discovery-mode="all"` (also reproduced with `annotated` + `@CdiComponent` on the view)
- Java: 21 (Temurin 21.0.9)
- Browser: Chrome on Windows 11 and Chrome on macOS 26.5.2
- No `@PreserveOnRefresh` on the view, any superclass, or any router layout; no router layout at all in the reproducer

## Minimal reproducer

```java
import com.vaadin.cdi.annotation.RouteScoped;
import com.vaadin.flow.component.AttachEvent;
import com.vaadin.flow.component.DetachEvent;
import com.vaadin.flow.component.html.Div;
import com.vaadin.flow.router.Route;
import jakarta.annotation.PostConstruct;
import jakarta.annotation.PreDestroy;

import java.util.UUID;

@Route("test")
@RouteScoped
public class TestView extends Div {

    private final UUID id = UUID.randomUUID();

    public TestView() {
        System.err.println("CONSTRUCTOR " + id);
    }

    @PostConstruct
    void init() {
        System.err.println("POST_CONSTRUCT " + id);
    }

    @PreDestroy
    void destroy() {
        System.err.println("PRE_DESTROY " + id);
    }

    @Override
    protected void onAttach(AttachEvent event) {
        String text = "ATTACH " + id + " UI " + event.getUI().getUIId();
        add(new Div(text));
        System.err.println(text);
    }

    @Override
    protected void onDetach(DetachEvent event) {
        System.err.println("DETACH " + id + " UI " + event.getUI().getUIId());
    }
}
```

## Steps to reproduce

Refresh variant (timing-dependent, matches the behavior we originally observed in production):

1. Open `/test`.
2. Press F5 repeatedly in quick succession. A handful of refreshes is often enough.
3. Eventually a refresh destroys the current instance (`DETACH` + `PRE_DESTROY`) and still uses it as the route target of the new UI — no new `CONSTRUCTOR` — and navigation fails with the same exception. Log excerpt from such a run (note the UI id jump caused by the rapid refreshes):

```
DETACH b0ff2469-c1a4-4245-9501-dec8507a14d6 UI 34
PRE_DESTROY b0ff2469-c1a4-4245-9501-dec8507a14d6
ERROR Can't move a node from one state tree to another. [...]
Offending component: created: Component 'TestView' at 'TestView.java' (<init> LINE 19), attached: [...]
```

A single calm F5 in a fresh, isolated tab does **not** trigger the problem — the old UI is cleaned up before the new UI registers its window name. This is why the issue looks browser- or machine-dependent: it follows the browser's `window.name` behavior (tab duplication, session restore) and request timing, not the browser itself.

## Actual result

The `@RouteScoped` bean of the first UI is resolved as the route target for the second UI. Stack trace (Vaadin CDI 16.1.0, Flow 25.2.4, no instrumentation):

```
java.lang.IllegalStateException: Can't move a node from one state tree to another. [...]
	at com.vaadin.flow.internal.StateNode.doSetTree(StateNode.java:835)
	at com.vaadin.flow.internal.StateNode.lambda$setTree$3(StateNode.java:401)
	at com.vaadin.flow.internal.StateNode.visitNodeTree(StateNode.java:771)
	at com.vaadin.flow.internal.StateNode.setTree(StateNode.java:401)
	at com.vaadin.flow.internal.StateNode.setParent(StateNode.java:283)
	at com.vaadin.flow.internal.nodefeature.NodeFeature.attachPotentialChild(NodeFeature.java:79)
	at com.vaadin.flow.internal.nodefeature.StateNodeNodeList.add(StateNodeNodeList.java:55)
	at com.vaadin.flow.internal.nodefeature.ElementChildrenList.add(ElementChildrenList.java:43)
	at com.vaadin.flow.dom.impl.AbstractNodeStateProvider.insertChild(AbstractNodeStateProvider.java:107)
	at com.vaadin.flow.dom.Node.insertChild(Node.java:394)
	at com.vaadin.flow.dom.Node.appendChild(Node.java:164)
	at com.vaadin.flow.component.internal.UIInternalUpdater.updateRoot(UIInternalUpdater.java:73)
	at com.vaadin.flow.component.internal.UIInternals.showRouteTarget(UIInternals.java:1023)
	at com.vaadin.flow.router.internal.AbstractNavigationStateRenderer.handle(AbstractNavigationStateRenderer.java:247)
	at com.vaadin.flow.component.internal.JavaScriptNavigationStateRenderer.handle(JavaScriptNavigationStateRenderer.java:80)
	at com.vaadin.flow.router.Router.executeNavigation(Router.java:572)
	at com.vaadin.flow.component.UI.handleNavigation(UI.java:2205)
	at com.vaadin.flow.component.UI.renderViewForRoute(UI.java:2166)
	at com.vaadin.flow.component.UI.browserNavigate(UI.java:2050)
	[...]
	at com.vaadin.cdi.CdiVaadinServlet.service(CdiVaadinServlet.java:65)
```
## Expected result

Without `@PreserveOnRefresh`, a `@RouteScoped` route target must never be shared between UIs. Each navigation in a new UI should create a fresh contextual instance, and an instance whose `@PreDestroy` has run must never be attached again:

```
DETACH old-instance
PRE_DESTROY old-instance
CONSTRUCTOR new-instance
POST_CONSTRUCT new-instance
ATTACH new-instance UI new-ui-id
```

This is also the documented behavior ("by default, Vaadin recreates all routing components, including @UIScoped and @RouteScoped beans, when a user refreshes the page", https://vaadin.com/docs/latest/flow/integrations/cdi/contexts).

mcollovati:
Route scope storage was keyed by window name whenever the extended client details reported one. That check used to be an implicit `@PreserveOnRefresh` detector: up to Vaadin 24, Flow requested the window name only for preserved navigation chains.

Since vaadin/flow#22719 the client details are collected during UI initialization and every tab gets a window name, so the check is always true. All `@RouteScoped` beans became shared between the UIs of the same browser window. When two such UIs are alive at the same time (duplicated tab, restored session, or reloads faster than the unload beacons that close the previous UI), the second UI is given a route target that still belongs to the first UI's state tree, and navigation fails with "Can't move a node from one state tree to another".

The storage is now bound to the window only when the navigation target or one of its layouts is annotated with `@PreserveOnRefresh`, mirroring Flow's isPreserveOnRefreshTarget rule. Otherwise it is bound to a single UI and destroyed on its detach.

Fixes #526

### DIFF
--- a/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
+++ b/vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
@@ -41,6 +41,7 @@ import com.vaadin.flow.component.UI;
 import com.vaadin.flow.component.page.ExtendedClientDetails;
 import com.vaadin.flow.router.AfterNavigationEvent;
 import com.vaadin.flow.router.BeforeEnterEvent;
+import com.vaadin.flow.router.PreserveOnRefresh;
 import com.vaadin.flow.router.RouterLayout;
 import com.vaadin.flow.server.VaadinSession;
 
@@ -54,6 +55,17 @@ public class RouteScopedContext extends AbstractContext {
     @VaadinSessionScoped
     public static class ContextualStorageManager
             extends AbstractContextualStorageManager<RouteStorageKey> {
+
+        /**
+         * Prefix of storage identifiers bound to a single UI instance.
+         */
+        private static final String UI_STORE_ID_PREFIX = "uid-";
+
+        /**
+         * Prefix of storage identifiers bound to a browser window, and thus
+         * shared by all the UIs of that window.
+         */
+        private static final String WINDOW_STORE_ID_PREFIX = "win-";
 
         public ContextualStorageManager() {
             // Session lock checked in VaadinSessionScopedContext while
@@ -92,10 +104,10 @@ public class RouteScopedContext extends AbstractContext {
 
         private void destroyDescopedBeans(UI ui,
                 Set<Class<?>> navigationChain) {
-            String uiStoreId = getUIStoreId(ui);
+            Set<String> uiStoreIds = getUIStoreIds(ui);
 
             Set<RouteStorageKey> missingKeys = getKeySet().stream()
-                    .filter(key -> key.getUIId().equals(uiStoreId))
+                    .filter(key -> uiStoreIds.contains(key.getUIId()))
                     .filter(key -> !navigationChain.contains(key.getOwner()))
                     .collect(Collectors.toSet());
 
@@ -103,6 +115,12 @@ public class RouteScopedContext extends AbstractContext {
         }
 
         private void handleUIDetach(UI ui, RouteStorageKey key) {
+            if (!key.isWindowScoped()) {
+                // The storage belongs to this UI only, so there is nothing to
+                // preserve for a potential UI created by a page refresh.
+                destroy(key);
+                return;
+            }
             UI uiAfterRefresh = findPreservingUI(ui);
             if (uiAfterRefresh == null) {
                 destroy(key);
@@ -131,36 +149,62 @@ public class RouteScopedContext extends AbstractContext {
         }
 
         private RouteStorageKey getKey(UI ui, Class<?> owner) {
-            ExtendedClientDetails details = ui.getInternals()
-                    .getExtendedClientDetails();
-            RouteStorageKey key = new RouteStorageKey(owner, getUIStoreId(ui));
-            if (details.getWindowName() == null) {
-                ui.getPage().retrieveExtendedClientDetails(
-                        det -> relocate(ui, key));
+            // Beans are shared with the UI created by a page refresh only if
+            // the navigation chain is preserved by Flow. In that case the
+            // storage is bound to the browser window, exactly like Flow binds
+            // the preserved component chain.
+            if (isPreserveOnRefreshChain(ui)) {
+                String windowName = getWindowName(ui);
+                if (windowName != null) {
+                    return new RouteStorageKey(owner,
+                            WINDOW_STORE_ID_PREFIX + windowName, true);
+                }
             }
-            return key;
-        }
-
-        private void relocate(UI ui, RouteStorageKey key) {
-            relocate(key,
-                    new RouteStorageKey(key.getOwner(), getUIStoreId(ui)));
+            return new RouteStorageKey(owner, getUIStoreId(ui), false);
         }
 
         private List<ContextualStorage> getActiveContextualStorages() {
-            return getKeySet().stream().filter(
-                    key -> key.getUIId().equals(getUIStoreId(UI.getCurrent())))
+            Set<String> uiStoreIds = getUIStoreIds(UI.getCurrent());
+            return getKeySet().stream()
+                    .filter(key -> uiStoreIds.contains(key.getUIId()))
                     .map(key -> getContextualStorage(key, false))
                     .collect(Collectors.toList());
         }
 
-        private String getUIStoreId(UI ui) {
-            ExtendedClientDetails details = ui.getInternals()
-                    .getExtendedClientDetails();
-            if (details.getWindowName() == null) {
-                return "uid-" + ui.getUIId();
-            } else {
-                return "win-" + getWindowName(ui);
+        /**
+         * Gets all the storage identifiers a UI can hold beans for: its own
+         * one, plus the one shared by all the UIs of the same browser window. A
+         * single UI may own both kinds of storage, for example after navigating
+         * from a regular view to a {@link PreserveOnRefresh} one.
+         */
+        private Set<String> getUIStoreIds(UI ui) {
+            Set<String> ids = new HashSet<>();
+            ids.add(getUIStoreId(ui));
+            String windowName = getWindowName(ui);
+            if (windowName != null) {
+                ids.add(WINDOW_STORE_ID_PREFIX + windowName);
             }
+            return ids;
+        }
+
+        private String getUIStoreId(UI ui) {
+            return UI_STORE_ID_PREFIX + ui.getUIId();
+        }
+
+        private static boolean isPreserveOnRefreshChain(UI ui) {
+            NavigationData data = ComponentUtil.getData(ui,
+                    NavigationData.class);
+            if (data == null) {
+                return false;
+            }
+            return isPreserveOnRefresh(data.getNavigationTarget())
+                    || data.getLayouts().stream().anyMatch(
+                            ContextualStorageManager::isPreserveOnRefresh);
+        }
+
+        private static boolean isPreserveOnRefresh(Class<?> clazz) {
+            return clazz != null
+                    && clazz.isAnnotationPresent(PreserveOnRefresh.class);
         }
 
     }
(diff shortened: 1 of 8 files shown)


vaadin-cdi/src/main/java/com/vaadin/cdi/context/RouteScopedContext.java
```diff
@@ -131,47 +149,77 @@ private static String getWindowName(UI ui) {
...
+                            WINDOW_STORE_ID_PREFIX + windowName, true);
+                }
             }
-            return key;
-        }
-
-        private void relocate(UI ui, RouteStorageKey key) {
-            relocate(key,
-                    new RouteStorageKey(key.getOwner(), getUIStoreId(ui)));
+            return new RouteStorageKey(owner, getUIStoreId(ui), false);
```
mshabarov: The following edge case deserves a look, what do you think? I don't know how practical it is though.

Edge case: scope is decided per navigation chain, not per owner — a layout-owned bean is recreated when navigating between a preserved and a non-preserved child 
  of the same layout

  The window-vs-UI scope here is chosen from isPreserveOnRefreshChain(ui), which looks at the whole current chain (target + all layouts). But RouteStorageKey
  equality is owner + uiId, so for one and the same owner the key flips between win-<windowName> and uid-<uiId> depending on which sibling view is currently
  active.

  Concretely: a shared RouterLayout without @PreserveOnRefresh, hosting one child with @PreserveOnRefresh and one without, plus a @RouteScoped 
  @RouteScopeOwner(SharedLayout.class) bean:

  1. Navigate to the preserved child → chain is preserve → the layout-owned bean is stored under (SharedLayout, "win-W").
  2. Navigate to the plain child (same layout instance is reused by Flow) → chain is no longer preserve → the bean is now looked up under (SharedLayout, "uid-B") →
  not found → a new instance is created (state lost), while the old (SharedLayout, "win-W") storage is never descoped (its owner is still in the chain) so it
  leaks until the UI detaches and @PreDestroy never runs.

  The transition from window-scope to UI-scope for a persistent owner is not migrated (this is roughly what the removed relocate(...) used to handle). I think the
  correct fix is to decide the scope per owner — i.e. whether this owner's position in the chain is preserved — rather than for the whole chain at once.

  It's a narrow combination (non-preserve shared layout + mixed preserve/plain children + a layout-scoped bean), and it doesn't affect the main bug this PR fixes,
  so it's fine to merge and track this separately. But it fails silently (no exception, no log), which makes it worth a follow-up.

  Reproducer (fails on this branch):

  // A shared layout WITHOUT @PreserveOnRefresh, reused across navigations
  public static class SharedLayout extends HasElementTestBean
          implements RouterLayout {
  }
  
  @RouteScoped
  @RouteScopeOwner(SharedLayout.class)
  public static class MemberOfSharedLayout extends HasElementTestBean {
      boolean isDestroyed;

      @PreDestroy
      private void onDestroy() {
          isDestroyed = true;
      }
  }

  @Inject
  @RouteScopeOwner(SharedLayout.class)
  private Provider<MemberOfSharedLayout> memberOfSharedLayout;

  @Test
  public void sharedLayout_navigateFromPreservedChildToPlainChild_layoutBeanIsKept() {
      doSetUp("foo", null);

      // 1) navigate to a @PreserveOnRefresh child under the shared layout
      Mockito.when(event.getNavigationTarget())
              .thenReturn((Class) PreservedGroup.class);
      Mockito.when(event.getLayouts()).thenReturn(
              Collections.<Class<? extends RouterLayout>> singletonList(
                      SharedLayout.class));
      beforeNavigationTrigger.fire(event);

      MemberOfSharedLayout bean1 = memberOfSharedLayout.get();
      bean1.setState(STATE);
  
      // 2) navigate to a plain (not preserved) child under the SAME layout
      // instance; the layout is reused, so its @RouteScoped bean must survive
      Mockito.when(event.getNavigationTarget())
              .thenReturn((Class) Group1.class);
      Mockito.when(event.getLayouts()).thenReturn(
              Collections.<Class<? extends RouterLayout>> singletonList(
                      SharedLayout.class));
      beforeNavigationTrigger.fire(event);

      MemberOfSharedLayout bean2 = memberOfSharedLayout.get();
      Assertions.assertFalse(bean1.isDestroyed,
              "The layout-owned bean must not be destroyed while the layout "
                      + "stays in the navigation chain");
      Assertions.assertSame(bean1, bean2,
              "The layout-owned bean must stay the same instance while the "
                      + "layout is reused across navigations");
      Assertions.assertEquals(STATE, bean2.getState());
  }

mcollovati: Good catch! I'll take a look at that.

mcollovati: This is a real problem, and it is a regression of this PR: on `main` the storage was always keyed by window, so the bean survived. I could reproduce it in a small app.

One note: choosing the scope per owner does not help. Flow's `isPreserveOnRefreshTarget` looks at the whole chain, so on refresh the full chain is preserved, also the layout without the annotation. This means the window binding is correct while the preserved child is active, and the UI binding is correct while the plain child is active. What is missing is the migration between the two, which is what the `relocate(...)` I removed was doing.

Fixed by moving the storage to the new key in both directions, instead of creating a second one. The move is skipped while another UI of the same window is still alive.

mshabarov (approved):

